PR #164, opened 2013-02-16
Added count aggregator.

adrien-mogenet:
What do you think of integrating a `count()` aggregator?

I don't have relevant use case as a global aggregator, but I think it makes senses when downsampling, sometimes. For example, if you're plotting query latencies, you might want to consider min/max/avg/dev over a time period, but you might also want to check the number of requests you've processed.


manolama:
Merged in 46c2b90c0c97163cd0712cd8240b2aa25b4897b7

PR #186, opened 2013-05-17
fix script to avoid glob expansion (which can exceed ARG_MAX)

tamsky:


manolama:
Merged in cd49dd05eca64ff1a3af1255c909dc04ae781c6f

PR #203, opened 2013-07-09
Migrate tool

sebastienc:
Hi guys,

At Hopper, we've been migrating from  one cluster to the other and wanted a cleaner way to migrate than using CopyTable. This so we could get rid of some legacy crap that's been lying around for a while. For this reason, I've tried to hack something together taking examples on DumpSeries and TextImport. It's not that optimized but with the amount of time I had to hack this together and with my current Java knowledge, it's the best I could do.

I hope it's not too bad!

Sebastien Coutu


manolama:
Gonna leave this in the archives since we'd really need a full MR type of job to migrate a lot of data. Thanks!

PR #243, opened 2013-10-04
Request for merge for change that ignores data points that utilize undefined tag values, tag keys, or metric names as opposed to throwing an exception (optional based on query string param)

davidkbainbridge:
## Overview

This change adds a query string parameter, 'silent'. If present then any query that contains a tag key, tag value, or metric name that has not been registered in TSDB is ignored (dropped) as opposed to throwing an exception. Thus an entire query will not fail if a member of it has one of these values that does not exist.
## Justification

A query that specifies tag information or a metric name does not need to fail and should instead return no data as this is essentially the same as specifying a filter that does not match any targets and thus should be treated as that way.

This is very helpful when you are using OpenTSDB to collect data and at time of query you can not determine if a given metric or tag pair has been added (--auto-metric). 

This is also helpful when a single query contains multiple data points some of which are valid (i.e. have valid tag / metric information) and some don't. Instead of the whole query failing, valid data is returned for those data points for which the filter matches targets.
## Backward Compatibility

This capability is activated by the presence of a 'silent' option on the end of the query string, i.e. '&silent'. Thus if this option is not include the existing behavior (exception) will be the default.

### DIFF
--- a/src/tsd/GraphHandler.java
+++ b/src/tsd/GraphHandler.java
@@ -854,26 +854,42 @@ final class GraphHandler implements HttpRpc {
       try {
         tsdbquery.setTimeSeries(metric, parsedtags, agg, rate);
       } catch (NoSuchUniqueName e) {
-        throw new BadRequestException(e.getMessage());
-      }
-      // downsampling function & interval.
-      if (i > 0) {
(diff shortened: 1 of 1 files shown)


manolama:
Partially implemented with the skip_unresolved_tags and even more options in v3. Thanks!

PR #254, opened 2013-11-11
add ability to check percentage of bad values

filippog:
rationale being that sometimes it might be OK if some points are outside the
threshold for a given window. e.g. test that _all_ points in the lookback
window must be outside the threshold, not just one.


manolama:
Merged in 1e07aa25bf99758cb654febce03d149c11282053

PR #265, opened 2013-12-23
Atomic increment/add

jakeri:
This is a new pull request for @KevinOrtman atomic increment feature (https://github.com/tsuna/opentsdb/pull/2) that was made against 1.X code base and never added master.

Basically the same code, except minor changes for 2.0 code base and using add-keyword instead of inc. There was some discussion about this in above mentioned pull request.

Another thing done is that I have disabled scheduleForCompaction(...) within the increment. Does this mean that a row created by add will never be compacted?
I would guess that compaction and atomic increment for already compacted values is not good fit.

Would like to see a discussion if this could be incorporated into the project.

### DIFF
--- a/src/core/TSDB.java
+++ b/src/core/TSDB.java
@@ -17,13 +17,11 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
-import com.stumbleupon.async.Callback;
-import com.stumbleupon.async.Deferred;
-import com.stumbleupon.async.DeferredGroupException;
+import net.opentsdb.stats.Histogram;
+import net.opentsdb.stats.StatsCollector;
+import net.opentsdb.uid.UniqueId;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
+import org.hbase.async.AtomicIncrementRequest;
 import org.hbase.async.Bytes;
 import org.hbase.async.ClientStats;
 import org.hbase.async.DeleteRequest;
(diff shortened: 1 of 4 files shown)


src/core/TSDB.java
```diff
@@ -32,10 +30,12 @@
 import org.hbase.async.HBaseException;
 import org.hbase.async.KeyValue;
 import org.hbase.async.PutRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import net.opentsdb.uid.UniqueId;
-import net.opentsdb.stats.Histogram;
-import net.opentsdb.stats.StatsCollector;
+import com.stumbleupon.async.Callback;
```
manolama: Avoid re-arranging imports if possible. Just clutters PRs.

src/core/TSDB.java
```diff
@@ -351,6 +351,75 @@ public WritableDataPoints newDataPoints() {
     return client.put(point);
   }
 
+  
+	/**
+	 * Increments a single integer value data point in the TSDB.
```
manolama: Need TSDB style formatting

src/core/TSDB.java
```diff
@@ -351,6 +351,75 @@ public WritableDataPoints newDataPoints() {
...
+		IncomingDataPoints.checkMetricAndTags(metric, tags);
+		final byte[] row = IncomingDataPoints
+				.rowKeyTemplate(this, metric, tags);
+		final long base_time = (timestamp - (timestamp % Const.MAX_TIMESPAN));
+		Bytes.setInt(row, (int) base_time, metrics.width());
+
+		//I guess that these points should never be scheduled for compaction.
+		//scheduleForCompaction(row, (int) base_time);
+		
+		final short qualifier = (short) ((timestamp - base_time) << Const.FLAG_BITS | flags);
```
manolama: Needs to support qualifier prefixes and milliseconds.

src/tsd/IncDataPointRpc.java
```diff
@@ -0,0 +1,135 @@
+package net.opentsdb.tsd;
```
manolama: Styling and would need to support HTTP writes

farito:
Hi, what happend with this PR ? was merged?

Issue #289, opened 2014-02-27
Configurable Scanner timeout to avoid OOM errors in TSDB


jarajapu:
While running a load test on the TSDB that involved a decent volume of calls to the /api/query end point (call rate of 1-2 calls/sec) for couple hours, we noticed that the JVM running TSDB threw an OOM and became extremely slow.  

We took a heap dump and analyzed the profile to notice several large data sets in memory (500K - 1.5M rows) related to queries that were taking as long as 600 - 700 seconds to be processed by asyncHBase. 

Here is a stack trace : 

19:09:20.370 WARN  [HashedWheelTimer.warn] - An exception was thrown by TimerTask.
java.lang.OutOfMemoryError: Java heap space
    at org.jboss.netty.buffer.HeapChannelBuffer.<init>(HeapChannelBuffer.java:42) ~[netty-3.6.2.Final.jar:na]
    at org.jboss.netty.buffer.BigEndianHeapChannelBuffer.<init>(BigEndianHeapChannelBuffer.java:34) ~[netty-3.6.2.Final.jar:na]
    at org.jboss.netty.buffer.ChannelBuffers.buffer(ChannelBuffers.java:134) ~[netty-3.6.2.Final.jar:na]
    at org.jboss.netty.buffer.ChannelBuffers.buffer(ChannelBuffers.java:121) ~[netty-3.6.2.Final.jar:na]
    at org.hbase.async.HBaseRpc.newBuffer(HBaseRpc.java:685) ~[asynchbase-1.5.0.jar:d543609]
    at org.hbase.async.MultiAction.serializeOld(MultiAction.java:254) ~[asynchbase-1.5.0.jar:d543609]
    at org.hbase.async.MultiAction.serialize(MultiAction.java:210) ~[asynchbase-1.5.0.jar:d543609]
    at org.hbase.async.RegionClient.encode(RegionClient.java:1144) ~[asynchbase-1.5.0.jar:d543609]
    at org.hbase.async.RegionClient.sendRpc(RegionClient.java:892) ~[asynchbase-1.5.0.jar:d543609]
    at org.hbase.async.RegionClient.periodicFlush(RegionClient.java:257) ~[asynchbase-1.5.0.jar:d543609]
    at org.hbase.async.RegionClient.access$000(RegionClient.java:89) ~[asynchbase-1.5.0.jar:d543609]
[...]

PR #290, opened 2014-02-27
Added support for timing out long-running queries (disabled by default)

jmangs:
Addresses feature request in #289

Problem Description:

Basically we had a Ruby script running against our DB looking for metrics to clean-up - I was running 8 concurrent requests to OpenTSDB. We managed to hit a few metrics that have millions of data points - basically it keeps retrieving data from HBase even though it's been running for 500-1000 seconds already.

The VM that this node was on ran out of memory so random parts starting throwing OOM errors without actually crashing the Deferred queries; we had 8 queries each occupying about 400-500MB of JVM memory. In our case, we'd rather have these queries fail so we don't lock up all of our workers running queries that will never finish in a reasonable time.

Feature:
What I did was to add a configurable timeout property that errors out ScannerCB() after a certain amount of time.

By default, this property `tsd.query.timeout` defaults to `-1` since that is the normal functionality in OpenTSDB. This results in `ScannerCB()` never timing out.

If the time spent in HBase exceeds this timeout then an exception is thrown

```java
if (timeout >= 0 && hbase_time > timeout) {
             throw new InterruptedException("Query timeout exceeded!");
}
```

Hopefully InterruptedException is acceptable. Right now you get "java.lang.RuntimeException: Shouldn't be here" from the timeout because there's no specific catch for this exception... which makes sense to me since it shouldn't be a common intended occurrence unless you put timeouts to 10ms or something.

I'd consider this a useful feature for production databases. We have many different teams using these metrics and there's a possibility someone could accidentally query something with millions of data points and eat up a lot of resources on a node. In my case, I ran into this problem while trying to search a wide date range for unused metrics (I was checking every single metric one by one)

### DIFF
--- a/src/core/TsdbQuery.java
+++ b/src/core/TsdbQuery.java
@@ -367,6 +367,7 @@ final class TsdbQuery implements Query {
       boolean seenAnnotation = false;
       int hbase_time = 0; // milliseconds.
       long starttime = System.nanoTime();
+      long timeout = tsdb.getConfig().getLong("tsd.query.timeout");
       
       /**
       * Starts the scanner and is called recursively to fetch the next set of
@@ -401,6 +402,10 @@ final class TsdbQuery implements Query {
              }
              scanner.close();
              return null;
+           }
+           
+           if (timeout > 0 && hbase_time > timeout) {
+             throw new InterruptedException("Query timeout exceeded!");
            }
            
            for (final ArrayList<KeyValue> row : rows) {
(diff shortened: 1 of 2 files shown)


manolama:
Merged in 0bdb855cfe1cfd1aeaad816f8e83255fb98d81a0. Note that this only handles long running scans whereas a lot of time is actually spent in the aggregation and serialization routines. We'll ultimately need timeouts there as well.

PR #291, opened 2014-02-28
Added support for deleting UID mappings

jmangs:
Right now there's no support built into the CLI tools to delete the UID mappings that show up when you run API commands like `suggest` and `query` - you can delete the data using `tsdb scan --delete` but the metrics will still come back in suggest and in queries (with no data if you delete it all). We have a bunch of old metrics which no longer have data, but they still show up in suggest - this becomes a pain in metrilyx because we can't show only pertinent metrics.

To address this, I've added a command called delete to `tsdb uid`:

`tsdb uid delete metrics sys.cpu.1` would delete the UID forward/reverse mappings for the specified UID. I'm not 100% sure how safe this is or how much data (if any) is left over but this let's you at least permanently remove it. Also, you may have to forcibly call `dropcaches` to get the metric to unregister.

Note: Deleting the UID *before* cleaning up it's data means you cannot use the built in CLI commands like scan on the metric anymore. You will have to clean it up in HBase yourself, so be warned.

Anyways, this is just meant to be an administrative tool for cleaning up garbage like rename is. If there are any caveats to this approach it would be good to know as well.

### DIFF
--- a/src/tools/UidManager.java
+++ b/src/tools/UidManager.java
@@ -206,7 +206,18 @@ final class UidManager {
         usage("Wrong number of arguments");
         return 2;
       }
-      return delete(tsdb.getClient(), table, idwidth, args);
+
+      try {
+        tsdb.getClient()
+            .ensureTableExists(
+                tsdb.getConfig().getString("tsd.storage.hbase.data_table"))
+            .joinUninterruptibly();
+
+        return delete(tsdb, table, idwidth, args);
+      } catch (Exception e) {
+        LOG.error("Unexpected exception", e);
+        return 4;
+      }
     } else if (args[0].equals("fsck")) {
       return fsck(tsdb.getClient(), table);
     } else if (args[0].equals("metasync")) {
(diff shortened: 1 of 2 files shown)


gsaray101:
Is this merged with the latest master? what is the actual command to actually remove the metric from the tsdb? scan --delete not working for me either. I still see the metric in uid table.

sidhhu:
@manolama Just looking through this. Well, if is it necessary to delete the tagk and tagv from the data table, then probably a better approach would be to do it on query time with compaction (similar way as how HBase major compaction works). But that would kill the read performance so it can be a option to delete or not.

manolama:
Cleaned up a bit and merge in 58ead0fc138a03606711f436e3a865d80aa86ad6

xmj:
When will this merge make it into a release?

manolama:
It's in 2.2RC1 already.

digrich:
How can we delete uid for a specific tag value of any giving metric? For e.g. If we have metrics like:

1.  app.rate.in app=a1 lp=lp1
2.  app.rate.in app=a2 lp=lp1
3.  app.rate.in app=a1 lp=lp2

How can I delete the uid of only first metric?

Issue #318, opened 2014-04-21
Introduce a counter to track the number of data points written


tsuna:
Currently we relied on `tsd.rpc.received type=put` to get the number of data points that a TSD received, however since the HTTP API allows batching multiple data points in one RPC, the count no longer reflects the number of data points.

We should introduce a new counter to accurately track the number of data points written.  Something like `tsd.datapoint.received` or whatever.

PR #323, opened 2014-04-23
Added datapoint counter

virtuallynathan:
as in issue #318

Not sure if this is the only file I have to change.


YiuTerran:
Why this pr not merged yet?

manolama:
Because this didn't place the counter in the ``TSD.addPoint()`` method which could be called by other classes. tehere is code for that now though, not sure where it came from.

PR #327, opened 2014-05-06
Add timeout to batch put.

jesse5e:
Batch put can now be called with with waitflush=timeout_in_seconds. The request
will complete when there are items with errors (http 400), when HBase errors
in /api/put will send http 500 errors (instead of 400), when all requests are
written to HBase (http 200) or after timeout_in_seconds (http 408).
Additionally, moved logging for all urls starting with /api/put calls debug.

NOTE: Submitted on behalf of third-party contributor:
  * Guenther Schmuelling <schmuell@pepperdata.com>

### DIFF
--- a/src/tsd/PutDataPointRpc.java
+++ b/src/tsd/PutDataPointRpc.java
@@ -16,14 +16,19 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
-
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import com.stumbleupon.async.Callback;
 import com.stumbleupon.async.Deferred;
 
 import org.jboss.netty.channel.Channel;
 import org.jboss.netty.handler.codec.http.HttpMethod;
 import org.jboss.netty.handler.codec.http.HttpResponseStatus;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


src/tsd/PutDataPointRpc.java
```diff
@@ -35,12 +40,106 @@
...
   private static final AtomicLong requests = new AtomicLong();
   private static final AtomicLong hbase_errors = new AtomicLong();
   private static final AtomicLong invalid_values = new AtomicLong();
   private static final AtomicLong illegal_arguments = new AtomicLong();
   private static final AtomicLong unknown_metrics = new AtomicLong();
+  private static final AtomicLong batch_timeouts = new AtomicLong();
+
+  /** timers to watch the requests */
+  private final static ScheduledExecutorService scheduler =
+      Executors.newScheduledThreadPool(2);
```
manolama: Why two threads and is that enough? We may want to make this configurable

manolama:
I'm not super fond of the scheduled thread pool but it may be the best way to timeout the RPCs right now. Also could you look at https://github.com/OpenTSDB/opentsdb/pull/371? I think it may be a bit more performant to wait on the group instead of each individual RPC.

jesse5e:
#371 <https://github.com/OpenTSDB/opentsdb/pull/371> looks simple, and
works for the purpose. However, it blocks a thread while waiting for
completion of put requests. It could easily limit the injection performance
of TSD by the number of netty threads not by the capacity of HBase or TSD
node. I believe our approach is much more efficient.

manolama:
I'll merge both into one kind of PR that won't block the thread but will use a shared HashedWheelTimer for the timeout (in addition to the new AsyncHBase timeouts).

manolama:
Re-worked and merged into 756dbda1a19ac7e0ff03a8534c74ed3bec3d9fa8. Thanks a ton!

jesse5e:
Thanks

Issue #83, opened 2011-10-27
./build.sh failed on archlinux


ghost:
When execute ./build.sh on archlinux, the following error occurred:

```
../build-aux/gen_build_data.sh src/BuildData.java net.opentsdb
Generating src/BuildData.java
  File "<stdin>", line 3
    print "timestamp=%d" % t;
                   ^
SyntaxError: invalid syntax
make: *** [src/BuildData.java] Error 1
```

PR #340, opened 2014-06-03
Use the versioned python command as PEP-394 suggests

luuse:
I'm on Arch and tried building opentsdb and ran into the same problem as in #83. I updated the command to python2 and the build works just fine. I'm aware that it's Arch that doesn't follow convention to have python 2 as the default but the command python2 should be more portable as it's suggested to use this in [this](http://legacy.python.org/dev/peps/pep-0394/) PEP.

> so python should be used in the shebang line only for scripts that are source compatible with both Python 2 and 3.


luuse:
#364 solves this equally well.

PR #349, opened 2014-06-19
The symbolic of static content should use relative path

chaotian:
The symbolic of the html and js are created in compiling time. The current implementation use absolute path. It means that the compiled binary could only place in the same absolute path with it in the compile machine. It is not convince for use and deploy. So I change the symbolic of static content form absolute path into relative path


manolama:
Sorry, coming back to this, how does it affect the packages (deb, rpm, bsd)? Could you test it under those and make sure it still works please? Thanks!

PR #351, opened 2014-06-23
Support Ganglia v30x and v31x protocols.

jesse5e:
- Opens an UDP port (8649) to receive Ganglia v30x and v31x protocol messages.
- Decodes Ganglia messages using gmetric4j and xdr.
- Inserts just numeric metrics to OpenTSDB.
- Added ganglia version, hostname, ip address, and optionally cluster name to
  the metric as tags.
- Ignored units.
- Ignored string metrics.
- Collected Ganglia RPC statistics.
- Ganglia support is disabled by default.


manolama:
Looks good but could you create a full repo for it similar to the plugins I was poking around with please? Thanks!

jesse5e:
Chris,

I'm guessing that you are suggesting Ganglia plugin source files have its
own new repository.

I'd like to use Maven to build a Ganglia plugin jar, but I cannot create
the dependency on OpenTSDB such as the following.

    <dependency>
      <groupId>net.opentsdb</groupId>
      <artifactId>opentsdb</artifactId>
      <version>2.1.0</version>
    </dependency>

It would be nice if you could upload OpenTSDB jars to either
mvnrepository.com or org.sonatype.oss.

johann8384:
@jesse5e did you happen to create a repository for this plugin? I can put it with mine if you want, and are comfortable with the GPLv3 license:  https://github.com/inst-tech/opentsdb-plugins

Also, the Maven dependency should work from here: https://mvnrepository.com/artifact/net.opentsdb/opentsdb/2.3.0

johann8384:
It didn't look like there were any changes to merge in once this plugin has it's own stand-alone home.

PR #353, opened 2014-07-01
Reduce the memory footprint with HBASE append support and randomize metric ids to distribute load evenly across region servers

rajeshal:
Current sequential metric ids may cause hot spots on same region servers. Random metric ids can shre the load almost equally across the region servers with pre-split regions


manolama:
Cleaning up and merging separately. Already got the random Metric UIDs into the Put branch.

PR #357, opened 2014-07-03
Downsample before and after aggregation.

jesse5e:
Supported downsampling aggregated results.
- Optional pre-downsampling query param is applied to the downsampling before
  aggregation. The query parameters should end with "-pre" like "1h-avg-pre".
  The default pre-downsampling could be applied if not specified.
- Optional post-downsampling query param is applied to the downsampling after
  aggregation. It uses the existing downsampling query param syntax.
  No post-downsampling will happen if not specified.
- The 0 second downsampling interval turns off the corresponding downsampling
  explicitly.
- A query to aggregate CPU metrics of the last hour with 10s
  pre-downsampling and 1 hour post-downsampling would be like:
    "/api/query?start=1h-ago&m=sum:10s-avg-pre:1h-avg:rate:sys.cpu.0"


manolama:
I may want to pull this in with a query language update since it's complicating the super basic query syntax. But we may merge it anyway. Let me play around with it a bit.

johann8384:
@manolama Is this covered in the 3.0 changes?

gnydick:
@jesse5e I think you're not understanding how downsampling is supposed to work 

http://opentsdb.net/docs/build/html/user_guide/query/aggregators.html?highlight=interpolation
> Downsampling
As mentioned above, interpolation is one means of handling missing data. But some users hate the fact that linear interpolation is a way of lying about the data because it generates phantom values. Instead one means of handling unaligned values is through downsampling. For example, if sources report a value every minute but they're skewed in time across that minute, provide a downsampling on 1 minute for every query across that source data. This will have the result of snapping the values to the same timestamp across each time series so that interpolation is mostly avoided. Interpolation will still occur when a downsampling bucket is missing a value.
>
>See Downsampling for details and examples on avoiding interpolation.

32 is the correct value, as downsampling snaps the values to the closest interval and then applies the aggregate function. Since there is only one datapoint close enough to hour 4 to get snapped to it, 32, you are getting just the average of 32. This is the expected behavior.

Wanting to see the **20** value isn't something meaningful. Who cares how many cores on average there are over 4 hours? I mean that in an objective way. If you're tracking the roll out, all you care about is the total at any point in time and possibly the rate of change.

Any situation where you would want to see an average of more than one datapoint, there will be more than one datapoint  close enough to the time interval to actually get aligned and aggregated.

I can't imagine a real world example where this new functionality would ever be needed.

manolama:
@johann8384 Yes this is now covered in 3.x since folks can put the downsampler in any order they like and have more than one.

PR #369, opened 2014-07-11
add datapoint counter

virtuallynathan:
trying again to add this as requested...


src/core/TSDB.java
```diff
@@ -569,6 +572,8 @@ public WritableDataPoints newDataPoints() {
     } else {
       v = Bytes.fromLong(value);
     }
+    collector.record("datapoints.received", datapoints_received, "type=all");
```
manolama: Don't collect this every time, just when the ``collectStats()`` method is called.

PR #371, opened 2014-07-16
Allow synchronous writes from the HTTP endpoint

waisbrot:
If the query-string specifies the flag "sync", then wait until all writes have actually occurred before returning.

Based on [this mailing list discussion](https://groups.google.com/d/msg/opentsdb/1gcuDmbaLhE/VL69lq9GTWIJ).

If anyone would like to suggest some pointers for writing tests to go along with this, I'd be happy to update.

I've tested this change against my own client where I was seeing the problem, and this appears to fix it.

### DIFF
--- a/src/tsd/PutDataPointRpc.java
+++ b/src/tsd/PutDataPointRpc.java
@@ -16,6 +16,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map.Entry;
 import java.util.concurrent.atomic.AtomicLong;
 
 import com.stumbleupon.async.Callback;
@@ -103,8 +104,11 @@ final class PutDataPointRpc implements TelnetRpc, HttpRpc {
     
     final boolean show_details = query.hasQueryStringParam("details");
     final boolean show_summary = query.hasQueryStringParam("summary");
+    final boolean synchronous_write = query.hasQueryStringParam("sync");
     final ArrayList<HashMap<String, Object>> details = show_details
       ? new ArrayList<HashMap<String, Object>>() : null;
+    final HashMap<IncomingDataPoint,Deferred<Object>> deferred_results
+      = new HashMap<IncomingDataPoint,Deferred<Object>>();
     long success = 0;
     long total = 0;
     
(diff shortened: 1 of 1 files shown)


src/tsd/PutDataPointRpc.java
```diff
@@ -103,8 +104,11 @@ public void execute(final TSDB tsdb, final HttpQuery query)
     
     final boolean show_details = query.hasQueryStringParam("details");
     final boolean show_summary = query.hasQueryStringParam("summary");
+    final boolean synchronous_write = query.hasQueryStringParam("sync");
     final ArrayList<HashMap<String, Object>> details = show_details
       ? new ArrayList<HashMap<String, Object>>() : null;
+    final HashMap<IncomingDataPoint,Deferred<Object>> deferred_results
+      = new HashMap<IncomingDataPoint,Deferred<Object>>();
```
manolama: Size the deferred map based on the # of DPs here and only initialize if the "sync" flag has been passed in.

src/tsd/PutDataPointRpc.java
```diff
@@ -140,11 +144,11 @@ public void execute(final TSDB tsdb, final HttpQuery query)
           continue;
         }
         if (Tags.looksLikeInteger(dp.getValue())) {
-          tsdb.addPoint(dp.getMetric(), dp.getTimestamp(), 
-              Tags.parseLong(dp.getValue()), dp.getTags());
+          deferred_results.put(dp, tsdb.addPoint(dp.getMetric(), dp.getTimestamp(),
```
manolama: Also check the "sync" flag here and only then add to the deferreds list.

src/tsd/PutDataPointRpc.java
```diff
@@ -169,6 +173,20 @@ public void execute(final TSDB tsdb, final HttpQuery query)
...
+          result.getValue().joinUninterruptibly();
+        } catch (Exception e) {
+          if (show_details) {
+            details.add(this.getHttpDetails(e.getMessage(), result.getKey()));
+          }
+          success--;
+        }
+      }
+    }
+
```
manolama: We need to avoid the join in here so you should do something like this:
```
    class SyncCB implements Callback<Object, ArrayList<Object>> {
      @Override
      public Object call(final ArrayList<Object> results) throws Exception {
        for (int i = 0; i < results.size(); i++) {
          if (results.get(i) instanceof Exception) {
            if (details != null) {
              final Exception e = (Exception)results.get(i);
              details.add(getHttpDetails(e.getMessage(), dps.get(i)));
            }
            // decrement success counter
          }
        }
        // break out the query.sendReply() code into a method or something that
        // can be called either from this callback or directly if the user doesn't
        // care about sync
        
        return null;
      }
    }
    
    if (sync) {
      Deferred.groupInOrder(deferreds).addCallback(new SyncCB());
    }
```

manolama:
Yeah, great idea for users who want the true sync behavior. It's a little trickier with the Telnet style but could be useful there as well.
Just one comment on using an async callback to confirm the puts and other than that add some unit tests please :)

waisbrot:
Thanks for the comments. I found the test-code to be pretty dense, which is why I was asking for some pointers (maybe link me to a similar test?).

manolama:
Re-worked and merged a bit in 756dbda1a19ac7e0ff03a8534c74ed3bec3d9fa8. Thanks a ton!

PR #376, opened 2014-07-22
allow override of all variables in /etc/default/debian

filippog:
hi,
I noticed some variables couldn't be overridden from /etc/default/debian, this should fix it!

thanks,
filippo


manolama:
Merged in d9a3e6821711c7be16fe73a69050f39aea212a71

PR #389, opened 2014-09-11
Fix #335 by pointing at tsdb.local in /usr/share/opentsdb/bin

### LINKED ISSUE #335: Documentation vs Implementation tsdb.local
http://opentsdb.net/docs/build/html/user_guide/cli/index.html

The documentation says: 

```
You might want to adjust JVM parameters, for instance to turn on GC activity logging or to set the size of various memory regions. In order to do so, simply set the variable JVMARGS in ./tsdb.local.
```

Created a file at `/usr/share/opentsdb/bin/tsdb.local` on `Ubuntu 12.04.4 LTS`. Nothing happens. Ended up overriding the `JVMARGS` in `/usr/share/opentsdb/bin/tsdb`

oozie:
For distribution packages, `$localdir` points at `/usr/share/opentsdb`. It's counterintuitive and inconsistent with the source version behavior where `./tsdb` expects `tsdb.local` to be in the same directory (./build).

https://github.com/OpenTSDB/opentsdb/issues/335

### DIFF
--- a/src/core/Query.java
+++ b/src/core/Query.java
@@ -22,7 +22,7 @@ import com.stumbleupon.async.Deferred;
 import net.opentsdb.uid.NoSuchUniqueName;
 
 /**
- * A query to retreive data from the TSDB.
+ * A query to retrieve data from the TSDB.
  */
 public interface Query {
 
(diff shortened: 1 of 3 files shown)


manolama:
Merged in a9e9b1f98c00e364fa7639f00707c1e680f7c69d

@jtakkala now that we have the config file, it would make more sense to use that than this old tsdb.local override.

PR #394, opened 2014-09-16
Adds HttpRpcPlugins that run on the main Netty loop.

jroyalty:
Introduces a new plugin type:  `HttpRpcPlugin`.  These are first class entities like the pre-existing `RpcPlugin`s though they have a separate lifecycle.  Further, unlike RPC plugins, the HTTP variety run on the main Netty loop along with built-in `HttpRpc` commands.  HTTP RPC plugins automatically placed beneath the **/plugin** base web path to keep them totally separate from build-in, core `HttpRpc` commands.  Even though these two HTTP RPC commands types are separate, both are dispatched by `RpcHandler` so a certain amount of common configuration can be safely shared; the CORS settings in particular.

HTTP RPC plugins can be loaded using the new `tsd.http.rpc.plugins` configuration parameter.  This is a comma-separated list of class names that extend the base `HttpRpcPlugin` class.

### DIFF
--- a/src/tsd/AbstractHttpQuery.java
+++ b/src/tsd/AbstractHttpQuery.java
@@ -33,6 +33,8 @@ import org.jboss.netty.handler.codec.http.QueryStringDecoder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import net.opentsdb.core.TSDB;
+
 /**
  * Abstract base class for HTTP queries.
  * 
@@ -60,13 +62,17 @@ public abstract class AbstractHttpQuery {
   private final DefaultHttpResponse response =
     new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
 
+  /** The {@code TSDB} instance we belong to */
+  protected final TSDB tsdb;
+  
   /**
    * Set up required internal state.  For subclasses.
    * 
    * @param request the incoming HTTP request
    * @param chan the {@link Channel} the request was received on
    */
-  protected AbstractHttpQuery(final HttpRequest request, final Channel chan) {
+  protected AbstractHttpQuery(final TSDB tsdb, final HttpRequest request, final Channel chan) {
+    this.tsdb = tsdb;
     this.request = request;
     this.chan = chan;
     this.method = request.getMethod();
--- a/src/tsd/AbstractHttpQuery.java
+++ b/src/tsd/AbstractHttpQuery.java
@@ -1,5 +1,5 @@
 // This file is part of OpenTSDB.
-// Copyright (C) 2010-2012  The OpenTSDB Authors.
+// Copyright (C) 2010-2014  The OpenTSDB Authors.
 //
 // This program is free software: you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License as published by
@@ -17,6 +17,8 @@ import java.nio.charset.UnsupportedCharsetException;
 import java.util.List;
 import java.util.Map;
 
+import com.google.common.base.Objects;
+
 import org.jboss.netty.buffer.ChannelBuffer;
 import org.jboss.netty.channel.Channel;
 import org.jboss.netty.channel.ChannelFuture;
@@ -30,8 +32,6 @@ import org.jboss.netty.handler.codec.http.HttpVersion;
 import org.jboss.netty.handler.codec.http.QueryStringDecoder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import com.google.common.base.Objects;
 
 /**
  * Abstract base class for HTTP queries.
--- a/src/tsd/AbstractHttpQuery.java
+++ b/src/tsd/AbstractHttpQuery.java
@@ -173,10 +173,34 @@ abstract class AbstractHttpQuery {
    * is to simply log the request info.  Can be overridden by subclasses.
    */
   public void done() {
-    if (LOG.isInfoEnabled()) {
-      final int processing_time = processingTimeMillis();
-      LOG.info(chan.toString() + " HTTP " + request.getUri() + " done in " + processing_time + "ms");
-    }
+    final int processing_time = processingTimeMillis();
+    logInfo("HTTP " + request.getUri() + " done in " + processing_time + "ms");
+  }
+  
+  /**
+   * Sends <code>500/Internal Server Error</code> to the client.
+   * @param cause The unexpected exception that caused this error.
+   */
+  public void internalError(final Exception cause) {
+    logError("Internal Server Error on " + request().getUri(), cause);
+    sendStatusOnly(HttpResponseStatus.INTERNAL_SERVER_ERROR);
+  }
+
+  /**
+   * Sends <code>400/Bad Request</code> status to the client.
+   * @param exception The exception that was thrown
+   */
+  public void badRequest(final BadRequestException exception) {
+    logWarn("Bad Request on " + request().getUri() + ": " + exception.getMessage());
+    sendStatusOnly(HttpResponseStatus.BAD_REQUEST);
+  }
+
+  /**
+   * Sends <code>404/Not Found</code> to the client.
+   */
+  public void notFound() {
+    logWarn("Not Found: " + request().getUri());
+    sendStatusOnly(HttpResponseStatus.NOT_FOUND);
   }
   
   /**
@@ -241,4 +265,31 @@ abstract class AbstractHttpQuery {
         .add("querystring", querystring)
         .toString();
   }
+  
+  // ---------------- //
+  // Logging helpers. //
+  // ---------------- //
+  
+  protected Logger logger() {
+    return LOG;
+  }
+
+  protected final void logInfo(final String msg) {
+    if (logger().isInfoEnabled()) {
+      logger().info(chan.toString() + ' ' + msg);
+    }
+  }
+
+  protected final void logWarn(final String msg) {
+    if (logger().isWarnEnabled()) {
+      logger().warn(chan.toString() + ' ' + msg);
+    }
+  }
+
+  protected final void logError(final String msg, final Exception e) {
+    if (logger().isErrorEnabled()) {
+      logger().error(chan.toString() + ' ' + msg, e);
+    }
+  }
+
 }
--- a/src/tsd/AbstractHttpQuery.java
+++ b/src/tsd/AbstractHttpQuery.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Map;
 
 import com.google.common.base.Objects;
+import com.stumbleupon.async.Deferred;
 
 import org.jboss.netty.buffer.ChannelBuffer;
 import org.jboss.netty.channel.Channel;
@@ -57,6 +58,10 @@ public abstract class AbstractHttpQuery {
 
   /** Parsed query string (lazily built on first access). */
   private Map<String, List<String>> querystring;
+  
+  /** Deferred result of this query, to allow asynchronous processing.
+   * (Optional.) */
+  protected final Deferred<Object> deferred = new Deferred<Object>();
   
   /** The response object we'll fill with data */
   private final DefaultHttpResponse response =
--- a/src/tsd/AbstractHttpQuery.java
+++ b/src/tsd/AbstractHttpQuery.java
@@ -38,7 +38,7 @@ import com.google.common.base.Objects;
  * 
  * @since 2.1
  */
-abstract class AbstractHttpQuery {
+public abstract class AbstractHttpQuery {
   private static final Logger LOG = LoggerFactory.getLogger(AbstractHttpQuery.class);
 
   /** When the query was started (useful for timing). */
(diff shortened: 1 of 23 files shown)


src/tsd/HttpQuery.java
```diff
@@ -114,10 +92,6 @@
   /** Deferred result of this query, to allow asynchronous processing.  */
   private final Deferred<Object> deferred = new Deferred<Object>();
 
-  /** The response object we'll fill with data */
-  private final DefaultHttpResponse response =
-    new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
-
   /** The {@code TSDB} instance we belong to */
   private final TSDB tsdb;
```
manolama: This could move to the abstract class since almost every RPC implementation would need the TSD.

src/tsd/AbstractHttpQuery.java
```diff
@@ -0,0 +1,337 @@
...
+import org.jboss.netty.handler.codec.http.HttpResponseStatus;
+import org.jboss.netty.handler.codec.http.HttpVersion;
+import org.jboss.netty.handler.codec.http.QueryStringDecoder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Abstract base class for HTTP queries.
+ * 
+ * @since 2.1
```
manolama: 2.2 now

src/tsd/AbstractHttpQuery.java
```diff
@@ -0,0 +1,337 @@
...
+ */
+public abstract class AbstractHttpQuery {
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractHttpQuery.class);
+
+  /** When the query was started (useful for timing). */
+  private final long start_time = System.nanoTime();
```
manolama: all of these should likely be protected so the implementer can access them.

src/tsd/AbstractHttpQuery.java
```diff
@@ -0,0 +1,337 @@
...
+  private final HttpMethod method;
+
+  /** Parsed query string (lazily built on first access). */
+  private Map<String, List<String>> querystring;
+  
+  /** The response object we'll fill with data */
+  private final DefaultHttpResponse response =
+    new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
+
+  protected AbstractHttpQuery(final HttpRequest request, final Channel chan) {
```
manolama: javadoc :)

src/tsd/AbstractHttpQuery.java
```diff
@@ -0,0 +1,337 @@
...
+   * Returns only the path component of the URI as a string
+   * This call strips the protocol, host, port and query string parameters
+   * leaving only the path e.g. "/path/starts/here"
+   * <p>
+   * Note that for slightly quicker performance you can call request().getUri()
+   * to get the full path as a string but you'll have to strip query string
+   * parameters manually.
+   * @return The path component of the URI
+   * @throws NullPointerException if the URI is null
+   * @since 2.0
```
manolama: These @sinces can all be stripped out of this file

src/tsd/HttpRpcPlugin.java
```diff
@@ -0,0 +1,113 @@
...
+ * 
+ * <p>Notes on multi-threaded behavior:
+ * <ul>
+ *  <li>Plugins are created and initialized <strong>once</strong> per instance
+ *    of the TSD.  Therefore, these plugins are effectively singletons.
+ *  <li>Plugins will be executed from multiple threads so the {@link #execute}
+ *    and {@link collectStats} methods <strong>must be thread safe</strong> 
+ *    with respect to the plugin's internal state and external resources.
+ * </ul>
+ * @since 2.1
```
manolama: since 2.2

src/tsd/HttpRpcPluginQuery.java
```diff
@@ -0,0 +1,53 @@
...
+import org.jboss.netty.channel.Channel;
+import org.jboss.netty.handler.codec.http.HttpRequest;
+
+import net.opentsdb.core.TSDB;
+
+/**
+ * Query class for {@link HttpRpcPlugin}s.  Binds together a request, its 
+ * owning channel, and reponse helpers.
+ * 
+ * @since 2.1
```
manolama: since 2.2

src/tsd/PipelineFactory.java
```diff
@@ -66,11 +66,11 @@
    * @throws Exception if the HttpQuery handler is unable to load 
    * serializers
    */
-  public PipelineFactory(final TSDB tsdb) {
+  public PipelineFactory(final TSDB tsdb, final RpcPluginsManager manager) {
```
manolama: Needs an override for backwards compatibility, should just take the TSD and instantiate a new RpcPluginsManager.

src/tsd/RpcHandler.java
```diff
@@ -73,8 +71,9 @@
    * @throws IllegalArgumentException if there was an error with the CORS domain
    * list
    */
-  public RpcHandler(final TSDB tsdb) {
+  public RpcHandler(final TSDB tsdb, final RpcPluginsManager pluginsManager) {
```
manolama: Same here, needs an override for API compatibility.

src/tsd/RpcHandler.java
```diff
@@ -207,78 +148,158 @@ private void handleTelnetRpc(final Channel chan, final String[] command) {
   }
 
   /**
-   * Finds the right handler for an HTTP query and executes it.
-   * Also handles simple and pre-flight CORS requests if configured, rejecting
-   * requests that do not match a domain in the list.
+   * Creates a query instance capable of handling the given request.
+   * @param tsdb
```
manolama: finish jdocs

src/tsd/RpcPluginsManager.java
```diff
@@ -0,0 +1,593 @@
...
+import net.opentsdb.stats.StatsCollector;
+import net.opentsdb.utils.JSON;
+import net.opentsdb.utils.PluginLoader;
+
+/**
+ * Manager for the lifecycle of <code>HttpRpc</code>s, <code>TelnetRpc</code>s,
+ * <code>RpcPlugin</code>s, and <code>HttpRpcPlugin</code>.  This is intended
+ * to be a singleton.
+ * 
+ * @since 2.1
```
manolama: 2.2

src/tsd/RpcPluginsManager.java
```diff
@@ -0,0 +1,593 @@
...
+  private ImmutableMap<String, HttpRpc> http_commands;
+  /** HTTP commands from user plugins. */
+  private ImmutableMap<String, HttpRpcPlugin> http_plugin_commands;
+  /** List of activated RPC plugins */
+  private ImmutableList<RpcPlugin> rpc_plugins = null;
+  
+  /** The TSDB that owns us. */
+  private TSDB tsdb;
+  
+  public RpcPluginsManager() {
```
manolama: Jdoc

src/tsd/RpcPluginsManager.java
```diff
@@ -0,0 +1,593 @@
...
+    telnet_commands = ImmutableMap.of();
+    http_commands = ImmutableMap.of();
+    http_plugin_commands = ImmutableMap.of();
+    rpc_plugins = ImmutableList.of();
+  }
+  
+  /**
+   * @param tsdb The parent TSDB object
+   */
+  public void initialize(final TSDB tsdb) {
```
manolama: We may be able to skip the initialization call here and just perform the init in the constructor since we should only ever have one instance of the rpc manager.

src/tsd/RpcPluginsManager.java
```diff
@@ -0,0 +1,593 @@
...
+import net.opentsdb.utils.PluginLoader;
+
+/**
+ * Manager for the lifecycle of <code>HttpRpc</code>s, <code>TelnetRpc</code>s,
+ * <code>RpcPlugin</code>s, and <code>HttpRpcPlugin</code>.  This is intended
+ * to be a singleton.
+ * 
+ * @since 2.1
+ */
+public final class RpcPluginsManager {
```
manolama: Heck, may even want to rename it just ``RpcManager``

src/tsd/RpcPluginsManager.java
```diff
@@ -0,0 +1,593 @@
...
+    if (tsdb.getConfig().hasProperty("tsd.http.rpc.plugins")) {
+      final String[] plugins = tsdb.getConfig().getString("tsd.http.rpc.plugins").split(",");
+      final ImmutableMap.Builder<String, HttpRpcPlugin> httpPluginsBuilder = ImmutableMap.builder();
+      initializeHttpRpcPlugins(mode, plugins, httpPluginsBuilder);
+      http_plugin_commands = httpPluginsBuilder.build();
+    }
+  }
+  
+  @VisibleForTesting
+  protected ImmutableList<RpcPlugin> getRpcPlugins() {
```
manolama: jdocs for all of the below :)

src/tsd/RpcPluginsManager.java
```diff
@@ -0,0 +1,593 @@
...
+  }
+  
+  public static void collectStats(final StatsCollector collector) {
+    RpcPluginsManager mgr = INSTANCE.get();
+    if (mgr != null) {
+      mgr.doCollectStats(collector);
+    }
+  }
+  
+  private void doCollectStats(final StatsCollector collector) {
```
manolama: May as well merge this into ``collectStats`` above, save the inlining cost.

test/tsd/TestRpcHandler.java
```diff
@@ -15,25 +15,29 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
-import static org.powermock.api.mockito.PowerMockito.mock;
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.anyString;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Matchers.*;
```
manolama: no wildcards please :)

src/tsd/HttpQuery.java
```diff
@@ -64,7 +57,7 @@
  * It makes it easier to provide a few utility methods to respond to the
  * requests.
  */
-final class HttpQuery {
+final class HttpQuery extends AbstractHttpQuery {
```
manolama: There are many more functions in here that can move into the base class. You could really just leave the built-in homepage bits in here and move everything else. Would that work for you?

src/tsd/RpcHandler.java
```diff
@@ -12,60 +12,58 @@
...
-  /** Commands we serve on the HTTP interface. */
-  private final HashMap<String, HttpRpc> http_commands;
   /** List of domains to allow access to HTTP. By default this will be empty and
    * all CORS headers will be ignored. */
   private final HashSet<String> cors_domains;
   /** List of headers allowed for access to HTTP. By default this will contain a
    * set of known-to-work headers */
   private final String cors_headers;
+  /** RPC plugins.  Contains the handlers we dispatch requests to. */
+  private final RpcPluginsManager plugins_manager;
```
manolama: just an rpc manager now

src/tsd/RpcHandler.java
```diff
@@ -207,78 +148,158 @@ private void handleTelnetRpc(final Channel chan, final String[] command) {
...
+  private AbstractHttpQuery createQueryInstance(final TSDB tsdb,
+        final HttpRequest request,
+        final Channel chan) 
+            throws BadRequestException {
+    final String uri = request.getUri();
+    if (Strings.isNullOrEmpty(uri)) {
+      throw new BadRequestException("Request URI is empty");
+    } else if (uri.charAt(0) != '/') {
+      throw new BadRequestException("Request URI doesn't start with a slash");
+    } else if (plugins_manager.isHttpRpcPluginPath(uri)) {
```
manolama: Don't use regex to compare the paths of every HTTP request. A number of users are pushing data with the ``/api/put`` endpoint so we don't want to slow down their writes. It may be quicker to explode the URI and see if it matches ``PLUGIN_BASE_WEBPATH`` or just see if it contains the plugin base.

src/tsd/RpcPluginsManager.java
```diff
@@ -0,0 +1,593 @@
...
+        "Invalid HttpRpcPlugin path %s. Path is equal to root.", testPath);
+    Preconditions.checkArgument(Strings.isNullOrEmpty(uri.getQuery()),
+        "Invalid HttpRpcPlugin path %s. Path contains query parameters.", testPath);
+  }
+
+  @VisibleForTesting
+  protected String canonicalizePluginPath(final String origPath) {
+    Preconditions.checkArgument(!(Strings.isNullOrEmpty(origPath) || origPath.equals("/")),
+        "Path %s is a root.", origPath);
+    String newPath = origPath;
```
manolama: no camel case please.

manolama:
Finally got to it, sorry for the wait! Looks good, just some minor tweaks/nit-picks and one important one about the regex matching. Let me know if you have time to make the tweaks or if you'd like me to go ahead and tweak then merge. Thanks!

jroyalty:
@manolama: Awesome!  Thanks for reviewing it!  I'll address these this week.

src/tsd/AbstractHttpQuery.java
jroyalty: They do have getters below.  Were you thinking protected (and non-final) so the references could be mutated?

manolama: protected and final for some just to avoid having publicly accessible accessors as we likely want to avoid letting other bits muck around with the channels and responses directly.

src/tsd/RpcPluginsManager.java
jroyalty: @manolama I'll just go ahead and make this a proper singleton.  Should be less confusing that way, I'd imagine.

src/tsd/AbstractHttpQuery.java
jroyalty: After digging in further, I believe these need at least a package-private getter.  There is code in `RpcHandler` and many existing `HttpRpc`s that use request, method, channel, and response.  Is that cool?  Or just leave 'em as is?

src/tsd/HttpQuery.java
jroyalty: I moved a bunch more methods but the send-related ones are a little harder to tease apart due to their reliance on getting the MIME type from the serializer.  I can try to abstract out the determination of MIME type and push that up into the base class, but I worry that would push too much modeling up...

manolama:
Merged in 6f9aa4b1f9fb7a514c3388e90cc4556a19b4a5bf

nickman:
Is this slated for the "next" branch ? It's a great addition and I could really use it.

manolama:
It's in the put branch which is the next next :) or 2.2

nickman:
Aw... man.....  That could be ages....

errrr... No offense....

PR #423, opened 2014-12-08
Added p50-75-95-99 exact and estimated aggregators with test

mxk1235:
i need the ability to extract percentile numbers from opentsdb, instead of computing them and storing ahead of time.

### DIFF
--- a/src/core/Aggregators.java
+++ b/src/core/Aggregators.java
@@ -16,10 +16,11 @@ import java.util.HashMap;
 import java.util.NoSuchElementException;
 import java.util.Set;
 
-import org.apache.commons.math.stat.descriptive.DescriptiveStatistics;
+import org.apache.commons.math3.stat.descriptive.rank.Percentile;
+import org.apache.commons.math3.stat.descriptive.rank.Percentile.EstimationType;
+import org.apache.commons.math3.util.ResizableDoubleArray;
 
 import com.google.common.base.Preconditions;
-import com.matttproud.quantile.Estimator;
 
 /**
  * Utility class that provides common, generally useful aggregators.
@@ -38,10 +39,6 @@ public final class Aggregators {
   
   /** Aggregator that sums up all the data points. */
   public static final Aggregator SUM = new Sum(
-      Interpolation.LERP, "sum");
-
-  /** Aggregator that counts up all the data points. */
-  public static final Aggregator COUNT = new Count(
       Interpolation.LERP, "sum");
 
   /** Aggregator that returns the minimum data point. */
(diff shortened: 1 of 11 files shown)


mxk1235:
ping?

thanks in advance!

src/core/Aggregators.java
```diff
@@ -332,4 +406,90 @@ public Interpolation interpolationMethod() {
     
   }
 
+  /**
+   * Percentile aggregator based on apache commons math implementation
+   */
+  private static final class Percentile implements Aggregator {
+    private final Float percentile;
```
manolama: Wouldn't hurt much to make this a double since that's what Apache wants.

src/core/Aggregators.java
```diff
@@ -332,4 +406,90 @@ public Interpolation interpolationMethod() {
...
+    private final String name;
+
+    public Percentile(final Float percentile, final String name) {
+      Preconditions.checkArgument(percentile > 0 && percentile <= 100.d, "Invalid percentile value");
+      this.percentile = percentile;
+      this.name = name;
+    }
+
+    public long runLong(final Longs values) {
+      DescriptiveStatistics stats = new DescriptiveStatistics();
```
manolama: Instead of ``DescriptiveStats``, use the ``Percentile`` class since it has less overhead. Try something like:
```
    final Percentile evalulator = new Percentile(percentile)
      .withEstimationType(estimation_type);
    final ResizableDoubleArray local_values = new ResizableDoubleArray();
    while(vs.hasNextValue()) {
      local_values.addElement(vs.nextDoubleValue());
    }
    evalulator.setData(local_values.getElements());
    return evalulator.evaluate();
```

The default for the ``P99``, ``P95`` etc. aggregators should be ``EstimationType.R_3`` since that will look at the entire set and pick out the closest value. For the ``EP99``, etc, maybe pick one of the other types from http://en.wikipedia.org/wiki/Quantile. Then we don't need to import Matt Proud's library until it makes it into Apache Commons or another well used repo.

src/core/Aggregators.java
```diff
@@ -332,4 +406,90 @@ public Interpolation interpolationMethod() {
...
+
+  }
+
+  /**
+   * Estimated percentile aggregator based on
+   * implementation of the Cormode, Korn, Muthukrishnan, and Srivastava algorithm
+   * for streaming calculation of targeted high-percentile epsilon-approximate
+   * quantiles.
+   */
+  private static final class EstimatedPercentile implements Aggregator {
```
manolama: See the note above about using one of the other Apache library estimates.

test/core/TestTsdbQuery.java
```diff
@@ -2466,7 +2466,42 @@ public void runMimMaxFloatOffset() throws Exception {
     }
     assertEquals(600, dps[0].size());
   }
- 
+  
+  @Test
+  public void runPercentiles() throws Exception {
```
manolama: We need tests for each type of new aggregator as well as making sure the CTors throw the proper exceptions if someone tries to hack the source and construct an aggregator with some funky value like -42.5 as the percentile. Aggregators need more UTs either way :(

src/core/Aggregators.java
```diff
@@ -141,6 +178,43 @@ public Interpolation interpolationMethod() {
     
   }
 
+  private static final class Count implements Aggregator {
```
manolama: Also take this guy out. I'll pull in #164 since that was the first implementation.

mxk1235:
will take a look..  thanks!

manolama:
Finally got to it, sorry for the wait. Take a look at the comments and let me know what you think. Any reason why we should use Matt Proud's implementation for estimation when we have the entire data set in memory? I think that's more for real-time processing where you can't keep all of the data points resident.

wattwood:
I'd love for this to be in place.

mxk1235:
thanks for the comments, i should be able to give you a new patchset in the next few hours.

the reason i don't always want to use apache implementation is that it is more memory heavy, however i'll take a look at the apache package again.

manolama:
Cool thanks! It actually seemed to be pretty performant when I was playing around with it yesterday even compared to a naive computation. Since we already have all of the data in memory before computation it makes sense to use the R3 method. If we get into streaming larger sets though and the other R* methods don't cut down memory enough, then we could look at the other sampling libraries.

mxk1235:
done, and updated relevant Makefile and pom.xml generation to include commons-math3 jar.

wattwood:
Can these be merged into 2.1?   mxk1235/opentsdb.git is showing version 2.0.1.

manolama:
This will go into 2.2.

Could you also sign the contrib agreement please?

mxk1235:
Sure, where can I do that?

manolama:
https://docs.google.com/spreadsheet/embeddedform?formkey=dFNiOFROLXJBbFBmMkQtb1hNMWhUUnc6MQ

mxk1235:
do you have an estimate when 2.2 release will be cut?

manolama:
Merged in d475d8490b14ecff15a1f78702734960aa05c01c

PR #424, opened 2014-12-10
Added <pluginManagement> tag to pom.xml.in to fix eclipse maven errors.

bryan4887:
Without this tag, Eclipse throws errors in the pom file like: `Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-antrun-plugin:1.7:run (execution: default, phase: process-resources`

### DIFF
--- a/pom.xml.in
+++ b/pom.xml.in
@@ -60,240 +60,241 @@
   <packaging>jar</packaging>
 
   <build>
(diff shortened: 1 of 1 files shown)


manolama:
Tried it again but it breaks the regular Maven build. To confirm a fix, run:
```
sh build.sh pom.xml
mvn compile
```

What error did Eclipse give you anway?

Issue #304, opened 2014-03-22
OpenTSDB leaking Sockets?


kylebrandt:
I've hit too many open files, digging into this:
```
[root@ny-devtsdb04 fd]# ps aux | grep opentsdb.conf
root      2076  3.8  1.5 5489036 260496 pts/0  Sl   Mar21  61:23 java -enableassertions -enablesystemassertions -classpath /usr/local/share/opentsdb/asynchbase-1.4.1.jar:/usr/local/share/opentsdb/guava-13.0.1.jar:/usr/local/share/opentsdb/jackson-annotations-2.2.3.jar:/usr/local/share/opentsdb/jackson-core-2.2.3.jar:/usr/local/share/opentsdb/jackson-databind-2.2.3.jar:/usr/local/share/opentsdb/log4j-over-slf4j-1.7.2.jar:/usr/local/share/opentsdb/logback-classic-1.0.9.jar:/usr/local/share/opentsdb/logback-core-1.0.9.jar:/usr/local/share/opentsdb/netty-3.6.2.Final.jar:/usr/local/share/opentsdb/slf4j-api-1.7.2.jar:/usr/local/share/opentsdb/suasync-1.4.0.jar:/usr/local/share/opentsdb/tsdb-2.0.0.jar:/usr/local/share/opentsdb/zookeeper-3.3.6.jar:/usr/local/share/opentsdb net.opentsdb.tools.TSDMain --port=4242 --staticroot=build/staticroot --cachedir=/tmp/tsd --zkquorum=ny-devtsdb01,ny-devtsdb03,ny-devtsdb04 --config /root/dev/opentsdb.conf
root     23362  0.0  0.0 103256   872 pts/0    S+   17:32   0:00 grep opentsdb.conf
[root@ny-devtsdb04 fd]# ls -l /proc/2076/fd | grep socket | wc -l
4049
[root@ny-devtsdb04 fd]# netstat -ap | grep 2076 | wc -l
6
[root@ny-devtsdb04 fd]# lsof -p 2076 | grep sock | head
java    2076 root   17u  sock                0,6      0t0 13372319 can't identify protocol
java    2076 root   49u  sock                0,6      0t0 13394689 can't identify protocol
java    2076 root   50u  sock                0,6      0t0 13394637 can't identify protocol
java    2076 root   51u  sock                0,6      0t0 13395575 can't identify protocol
java    2076 root   52u  unix 0xffff880437c5a0c0      0t0 12977998 socket
[...]

PR #439, opened 2015-02-06
Updated I/O worker to handle IdleStateEvent after Netty upgrade. (#304)

jmangs:
Looks like for Netty 3.5.9 and later, the `IdleStateEvent` is now handed to the I/O worker thread. Was changed here [#641](https://github.com/netty/netty/issues/641) - this causes OpenTSDB to not close sockets after an idle period anymore since when I made the original change in #379 it was done on the Netty Timer Thread ([ConnectionManager](https://github.com/OpenTSDB/opentsdb/blob/next/src/tsd/ConnectionManager.java#L129) (this worked because I was using an older version of Netty at the time)

I've moved the handling of IdleEvents to `RpcHandler` where it should be now since we're on Netty 3.9.x.

Changes:
- Switched `ConnectionManager` back to `SimpleChannelHandler` since Timer Threads don't handle `IdleStateEvent` anymore
- Moved `IdleStateHandler` inside `ChannelPipeline` in the `PipelineFactory` class.
- `RpcHandler` now extends `IdleStateAwareChannelUpstreamHandler` and overrides `channelIdle` to handle idle sockets.

If you have DEBUG level logging on, you will see this if you have a timeout set:

```
2015-02-06 08:59:59,213 DEBUG  [New I/O worker #1] RpcHandler: Closed idle socket.
```

### DIFF
--- a/src/tsd/ConnectionManager.java
+++ b/src/tsd/ConnectionManager.java
@@ -23,19 +23,16 @@ import org.jboss.netty.channel.ChannelEvent;
 import org.jboss.netty.channel.ChannelHandlerContext;
 import org.jboss.netty.channel.ChannelStateEvent;
 import org.jboss.netty.channel.ExceptionEvent;
+import org.jboss.netty.channel.SimpleChannelHandler;
 import org.jboss.netty.channel.group.DefaultChannelGroup;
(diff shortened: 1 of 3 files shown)


manolama:
Merged with 9621c31f8e726deed2d76c9ebc1933c1ca8700fc

PR #440, opened 2015-02-07
Remove compactionq flush call from tsdb flush

nitin-aggarwal:
Flushing compaction queue from tsdb flush, is creating an endless
chain of compaction queue flushes, as compaction queue flush
calls tsdb flush to flush HBaseClient.

Also we don't really need to call compaction queue flush from tsdb.flush(),
as compaction background thread flush anyways every FLUSH_INTERVAL seconds.

### DIFF
--- a/src/core/CompactionQueue.java
+++ b/src/core/CompactionQueue.java
@@ -189,7 +189,7 @@ final class CompactionQueue extends ConcurrentSkipListMap<byte[], Boolean> {
     if (nflushes == MAX_CONCURRENT_FLUSHES && maxflushes > 0) {
       // We're not done yet.  Once this group of flushes completes, we need
       // to kick off more.
-      tsdb.flush();  // Speed up this batch by telling the client to flush.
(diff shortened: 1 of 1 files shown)


nitin-aggarwal:
This looks like the main reason for super fast compactions, which is leading to lot of issues with TSDs and RegionServers, running with good workload and bigger compaction queues.

manolama:
Hmm interesting. How about changing the code in CompactionQueue to ``tsdb.getClient().flush();`` so we don't break the public API of the TSD. Some users may still rely on flush calling into the compaction queue so I'd like to avoid breaking their apps. thanks!

nitin-aggarwal:
I initially thought about that, but I couldn't think of a good reason to start asynchronous compactions on tsdb flushes, if there is already a background thread doing compactions.
It is needed on tsdb close, as we can miss compacting some of the time-series uids there.

I have uploaded change to incorporate suggested change, to avoid API semantic changes.

manolama:
Merged in eecd448e93c3a0422cfdbe2ffcf5cf468b02dec6

PR #441, opened 2015-02-07
Make compaction parameters configurable

nitin-aggarwal:
This would help tune compaction to be more staggered, based on workloads.


manolama:
Merged in 57b546414c3f28c5167116a55f06e87fe9cbba93

PR #442, opened 2015-02-10
sort the aggregators in the web form

mxk1235:


manolama:
Merged in 6dc59e30383d406213fdffc114d60da5592a20a5

oozie:
There is no build problems, but `java.lang.ClassCastException` is thrown at runtime. 
I see two problems with this change:
- It was not tested. At all.
- Out of 5 lines, one introduces an unused import.
Please make sure to include testing results in your future pull requests.

mxk1235:
thanks for fixing this.  for some reason i thought it's covered by a test, but in fact the entire .client. package is excluded from the compiler plugin, and the regression tests didn't hit the page.

unfortunately, i am not familiar enough with the build process to modify any of the <exclude> sections.

PR #447, opened 2015-02-23
Only remove config files if uninstalling last copy

jtakkala:
The RPM post and postun scripts have weird behaviour when you're
upgrading. We should check the first argument to each script, to detect
if this is a fresh install, ugprade, or uninstall. See
http://www.rpm.org/max-rpm/s1-rpm-inside-scripts.html for an
explanation.

### DIFF
--- a/opentsdb.spec.in
+++ b/opentsdb.spec.in
@@ -81,13 +81,18 @@ rm -rf %{buildroot}
 
 %post
 
-ln -s %{_datarootdir}/opentsdb/etc/opentsdb /etc/opentsdb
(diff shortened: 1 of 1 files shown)


manolama:
Looks good, could you sign the contrib agreement please? https://docs.google.com/spreadsheet/embeddedform?formkey=dFNiOFROLXJBbFBmMkQtb1hNMWhUUnc6MQ Thanks!

jtakkala:
Already signed it years ago. Has it changed since then?

jtakkala:
Sorry to re-open this, but I notice that this PR isn't in the 'put' branch. Can you merge it into that? Do you need me to open another PR?

Issue #396, opened 2014-09-30
/tmp/opentsdb directory erased on a recurring basis, leading to GUI being unavailable


jletroui:
Regularly, GUI on one of the nodes is not working anymore and is producing the following error:
![tsd_and_1__default__bash_](https://cloud.githubusercontent.com/assets/450475/4458118/c226bd0e-489f-11e4-9bc3-c79b33b4f6f3.png)

Looking on the server, the /tmp/opentsdb has indeed disappeared. Recreating it manually is fixing the issue (temporarily).

This is happening approximately 0.1 time per week per user per node. Up to now, we have not identified a pattern to reproduce it solidly.

oozie:
It's `dumpToFiles`. Given that the default location for cachedir is `/tmp/opentsdb`, the right thing to do here is to defend against this by trying to recreate cache dir and log a warning on a failure.
```
Request failed: Internal Server Error
    java.io.FileNotFoundException: /tmp/opentsdb/e024de37_0.dat (No such file or directory)
    at java.io.FileOutputStream.open(Native Method) ~[na:1.7.0_65]
    at java.io.FileOutputStream.<init>(FileOutputStream.java:221) ~[na:1.7.0_65]
    at java.io.FileOutputStream.<init>(FileOutputStream.java:110) ~[na:1.7.0_65]
    at java.io.PrintWriter.<init>(PrintWriter.java:184) ~[na:1.7.0_65]
    at net.opentsdb.graph.Plot.dumpToFiles(Plot.java:199) ~[tsdb-2.1.0RC1.jar:36fba94]
    at net.opentsdb.tsd.GraphHandler.runGnuplot(GraphHandler.java:719) ~[tsdb-2.1.0RC1.jar:36fba94]
    at net.opentsdb.tsd.GraphHandler$RunGnuplot.execute(GraphHandler.java:305) ~[tsdb-2.1.0RC1.jar:36fba94]
    at net.opentsdb.tsd.GraphHandler$RunGnuplot.run(GraphHandler.java:292) ~[tsdb-2.1.0RC1.jar:36fba94]
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145) [na:1.7.0_65]
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615) [na:1.7.0_65]
    at java.lang.Thread.run(Thread.java:745) [na:1.7.0_65] 
`

[...]

PR #451, opened 2015-02-28
Address #396 to alleviate the recurring cachedir removal problem

oozie:
To reproduce:
* pull up a graph in TSD,
* `rm -fr /tmp/opentsdb`
* go back to the browser, press `F5`

With this change, the graph will load after `F5` is pressed.

### DIFF
--- a/test/graph/TestPlot.java
+++ b/test/graph/TestPlot.java
@@ -1,0 +1,97 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2011-2012  The OpenTSDB Authors.
+//
+// This program is free software: you can redistribute it and/or modify it
+// under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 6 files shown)


test/graph/TestPlot.java
```diff
@@ -0,0 +1,97 @@
+// This file is part of OpenTSDB.
```
manolama: Make sure to add this file to the Makefile so the sh script can test it

manolama:
Looks good. It might be worth it to move the ``checkDirectory()`` method to a Utility class like ``FileUtility.java``.

oozie:
Thanks for the review - suggestions applied.

manolama:
Merged in 8f06f6ecefefc8d187c3d29b81fa62fe4d0d2169

PR #456, opened 2015-03-06
Fix flaky TestUID.java leaking Netty threads

oozie:
Ineffective HBaseClient mock caused each subsequent test to leak 9 threads for netty boss and 8 workers. The image illustrates an improvement that should make this test more robust. 
thread_leaks
There is still a single thread leak per test, but who cares.


manolama:
I care!! heh, just kidding, I'll kill that too eventually but thanks a ton!

manolama:
Merged in dad3f7feeb6ae8d1b0fbf21f6481d9bc80fa8a6f

Issue #457, opened 2015-03-06
Global annotations retrieval in query behaving strangely


vitorboschi:
A few minutes after adding a global annotation, I can retrieve it with a GET call do /api/query with start=10m-ago, for example. However, if I use the exact same query with start=60m-ago, they don't show up.
This is the query I'm using:
```
/api/query?m=avg:1h-avg:slt.reader.temperature{readerId=90}&start=26m-ago&global_annotations=true&show_tsuids=true
```
The result (no annotations in this response)
```
[{"metric":"slt.reader.temperature","tags":{"host":"xxxxxx","sltId":"xx","readerId":"90"},"aggregateTags":[],"tsuids":["00000100000100004F00000200009900000300009B"],"dps":{"1425639600":48.08695652173913}}]
```
I've just discovered that if current time is XX:YY, whenever my start time is ZZm-ago, I receive the annotations when ZZ < YY, and not receive when it's more than YY (of course, assuming both ranges contains annotations).

In the example query, I'll not receive annotations until time reaches XX:26. Before that, it'll not return the annotations.

It looks like I can reproduce this isse whenever there's recent annotations.

Version info: OpenTSDB version [2.1.0RC1] built from revision a77d839 in a MINT state

oozie:
Nice catch and good job with the heuristics. It's a simple bug to fix but it's slightly harder to prove with a unit test, since in tests the annotations from the stop row are also pulled in (tests need some deeper fixing here).

Anyway, the offending lines are https://github.com/OpenTSDB/opentsdb/blob/master/src/meta/Annotation.java#L330-L333 -
So using the XX:YY and ZZ analogy, when you're ZZm-ago start time is greater than number of minutes in a given hour YY, the start row for the hbase scan at an hour before XX and stop row ends with the present hour (data from it is not being considered) - this is when the bug manifests itself.
When XX moves on to be the next hour, the start row falls on hour XX-1, but the stop row is your present hour XX+1, i.e. the scan for annotations is performed on rows XX-1 and XX+0, (XX+1 is skipped as the end row) here the annotation gets retrieved properly, which is why it looks like you can reproduce the issue with "recent" annotations only, where recent means annotations from the current hour of the day.

manolama:
Should be fixed with @oozies commit in RC2.

PR #458, opened 2015-03-08
Fix for #457 improper global annotation retrieval + a MockBase fix

oozie:
This change fixes improper global annotation retrieval as reported in #457. This bug is elusive in that it only manifests itself for a portion of every current hour for which annotations are being retrieved. See the original issue for more explanation.

The change is broken down into three commits:
- Fix of MockBase to reflect the actual Scanner's behavior - this commit breaks ~9 tests,
- Actual bugfix in `Annotation.java` to properly set the stop key correctly + disabling of a broken test with a TODO (this change fixes the 8 broken tests and disables the remaining one).
- A new unit test proving that the fix works.


manolama:
Merged in 10ac615bae0b249c232eda1c375aacc1e4b955fc

Issue #459, opened 2015-03-08
A likely bug in Tree.deleteTree()


oozie:
The test `TestTree.deleteTree()` had been passing under a false assumption about how asynchbase's Scanner works: `MockBase` has been including the stop row in the result set, which is not what happens in reality. `MockBase`'s behavior has been corrected in #458 but the test is now failing for either one of two reasons:
* it's a broken test with improper setup or 
* it surfaces an off-by-one bug in `Tree.deleteTree()`

Not sure which one it is; here is a TODO:
https://github.com/oozie/opentsdb/commit/db201c43c4d99cf3309100d3c20c9a99b3ba8048#diff-e95c7bedd8fc8751e18c663aa4fbef37R557

manolama:
Fixed with @Oozie's commit.

PR #461, opened 2015-03-11
Restoring TestTree.deleteTree() disabled in  #458

oozie:
Okay, so test mystery solved: Old `MockBase` code was going one key too far and mistakenly removing Tree 2. Despite this not being the objective of `TestTree.deleteTree()`, the test asserted this behavior as correct (while in actuality it's not). Therefore when `MockBase` got fixed, this test started failing as it operated on the faulty assertion.

### DIFF
--- a/test/tree/TestTree.java
+++ b/test/tree/TestTree.java
@@ -75,10 +75,7 @@ public final class TestTree {
   @Before
   public void before() throws Exception {
     final Config config = new Config(false);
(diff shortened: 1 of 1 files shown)


oozie:
This change addresses #459

manolama:
Merged in 9861051aa62b088d99dab8d6577df1c018ce4ae2

Issue #460, opened 2015-03-10
Add option to ignore Counter Reset 0 datapoint 


frogmaster:
I have a  counter, that wraps once per day, which leaves me with graph having drops to zero once per day. It would be ok, if the datapoint is discarded instead or sent with null value instead. Handling the counter wrap value "0" client side is quite hard, because there is no distinction between real zero and wrap zero...

kylebrandt:
I also wish was dropped. Having a zero can mess up reductions to a single number when using something like mean()

luuse:
What kind of metrics do you have that are represented by counters (are they strictly increasing or can they decrease)? When you graph them do you still look at them as counters or as rates?

frogmaster:
Mine specifically are interface octet counters, but from a rather nasty device, thus the daily wrap. They are strictly increasing, and i use rate.

manolama:
Good idea, it's a little tricky to implement but doable.

luuse:
I'm not entirely sure where you would like to drop the data point? Would you like it to be dropped at query time or not added to the database at all?

I haven't looked at the query code in a while and i'm still not sure if i understand how the graphs look currently. But i think to me the best solution would be to introduce an optional period parameter to the rate keyword. That should allow us to recognize when a counter rolls over as long as `v2 < v1` and continue the rate calculations as usual while still exposing real zeros?

kylebrandt:
When a counter resets, it is impossible to now how much incremented. For
example observation A is 1000, and observation B is 14. We don't know much
the value increased after observation and the reset. Therefore we can't
take the derivative since we are missing that information.

So when doing a "rate" query, a more proper result for the derivative would
NaN. Since opentsdb doesn't support NaN the best thing would be to drop it.
This is because the resultant 0 calculation of the rate is lie (The rate
wasn't 0, OpenTSDB just can't calculate it).

The 0 behavior is particularly undesirable  when doing the reduction of a
time series generate by rate to a single number (i.e. avg() ) because it
throws the avg off.

luuse:
So the problem you have is with counter resets and not how counter rollovers are currently handled? If so, does a properly configured `resetValue` correctly identify your resets? Finally, if `resetValue` identifies the resets, is it the `0` value that is currently returned when `resetValue` signals what you'd like to change?

frogmaster:
Yes, that exactly what i need. insead of returning 0 value, return Null/Nan, or better yet, drop the particular datapoint from query results.

[...]

PR #465, opened 2015-03-15
Address #460 by handling counter rollovers automatically

oozie:
rate_counter

This change detects resets of strictly increasing counters and skips the low-quality data point rather than fudging it with constant values. The change also gets rid of two redundant fields in the UI that consumed a lot of real estate and provided dubious value.

### DIFF
--- a/src/core/RateSpan.java
+++ b/src/core/RateSpan.java
@@ -157,43 +157,40 @@ public class RateSpan implements SeekableView {
     return difference;
   }
 
+  /*
+   * Marks the end of rate time series.
+   */
+  private void markEndOfSeries() {
+    // Invalidates the next rate with invalid timestamp.
+    next_rate.reset(INVALID_TIMESTAMP, 0);
+  }
   /**
    * Populate the next rate.
    */
   private void populateNextRate() {
     if (source.hasNext()) {
       moveToNextDatapoint();
+      double difference = getValueDifference();
+      if (options.isCounter() && difference < 0) {
+        while(source.hasNext()) {
+          moveToNextDatapoint();
+          difference = getValueDifference();
+          if (difference > 0) {
+            break;
+          }
+        }
+        if (!source.hasNext()) {
+          markEndOfSeries();
+          return;
+        }
+      }
       // TODO: for backwards compatibility we'll convert the ms to seconds
       // but in the future we should add a ratems flag that will calculate
       // the rate as is.
       final double time_delta_secs = getTimeDeltaSecs();
-      double difference = getValueDifference();
-
-      if (options.isCounter() && difference < 0) {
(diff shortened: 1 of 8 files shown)


src/core/RateSpan.java
```diff
@@ -116,61 +116,81 @@ private void initializeIfNotDone() {
...
-          // while converting long values to double values if both values are long.
-          difference = options.getCounterMax() - prev_data.longValue() +
-              next_data.longValue();
-        } else {
-          difference = options.getCounterMax() - prev_data.toDouble() +
-              next_data.toDouble();
+        while(source.hasNext()) {
+          moveToNextDatapoint();
+          difference = getValueDifference();
+          if (difference > 0) {
```
frogmaster: I think this should be >= 0, currently this just skips until counter increases, which it may not, for some time.

oozie: You're 100% correct.

kylebrandt:
What is blocking the merge on this at this point?

kylebrandt:
@manolama Any chance on getting this merged?

manolama:
Looking at it now. One thing is that we already have the RateOptions in the main 2.0 release so we can't get rid of those features as some folks may be using them. Instead this PR should add an auto-detect option to fix the rates but let folks override them if necessary.

oozie:
`RateOptions` as a way of conveying rate config stays. It seems fine and can be used for something useful in the future, e.g. defining custom interval (per-minute, per-host as opposed to just second).

I know you're referring to the two fields specifically, but let's at least evaluate the decision to drop them. The functionality they provide is so weak that I cannot conceive of anybody preferring it over a dropped datapoint. I'd like to be proven wrong, I spent hours thinking about it and concluded that people are more likely to thank you for dropping over keeping them.

Could we drop them for a test and I'll commit to adding them back as soon as somebody complains?

kylebrandt:
I'd love to see this soon. In Bosun we use the query string syntax in our alerts, i.e:

`q("sum:10m-avg:rate{counter,,1}:haproxy.logs.hits_by_route{route=$route}", "5m", "")`

Will that syntax stop working? Also what will the new syntax be (or are there no changes there?)

oozie:
@kylebrandt I just checked a `rate{counter,,1}` query and it returns the same set of data points as `rate{counter}` - a reason to believe it's backwards compatible.

manolama:
Yeah, dropping the fields from the UI is fine, but since users may have integrated the Java API and query strings we need to make sure those are still supported for at least a couple of minor versions. I'll go ahead and tweak your pull and get it in.

manolama:
Re-worked it into 4be59101e3797842c5be89044f2a186ee13ed9e3 so that we maintain the previous functionality but allow for dropping.

Issue #437, opened 2015-02-02
Make MAX_NUM_TAGS Configurable; Throw error when attempting to write data when tags excceds MAX_NUM_TAGS.


kylemallory:
First, there needs to be some kind of error or other indication when attempting to write a metric with more than 8 tags.  Currently (based on my log level), tsdb quietly drops the metric, and I'm left wondering why my metric isn't showing up in the database.

Secondly, could MAX_NUM_TAGS become a configurable/tunable item?  I have a few metrics which are pushing the tag limit.  I understand the performance reasons/concerns associated with changing this, but would rather not have to recompile/deploy everytime I tweak this value.

manolama:
I'll double check but it *should* respond to the sender that there were too many tags. Are you writing over the Telnet style interface? If so, check your receive buffer for errors. Those aren't logged right now because it can potentially fill up the disk if there are a lot of data points coming back with bad values.

And yeah, this can easily become a config value so we'll get that in (or if you have time to work on a patch... hint hint :)

nickman:
That could be ugly if you configure max-num-tags to a value lower that the actual max number of tags in hbase. Could you store the max in hbase somewhere and then check when OpenTSDB connects. It would allow higher, but not lower.

jtamplin:
On Mon, Feb 2, 2015 at 2:34 PM, kylemallory <notifications@github.com>
wrote:

> First, of all, there needs to be some kind of error or other indication
> when attempting to write a metric with more than 8 tags. Currently (based
> on my log level), tsdb quietly drops the metric, and I'm left wondering why
> my metric isn't showing up in the database.
>
> Secondly, could MAX_NUM_TAGS become a configurable/tunable item? I have a
> few metrics which are pushing the tag limit. I understand the performance
> reasons/concerns associated with changing this, but would rather not have
> to recompile/deploy everytime I tweak this value.
>
This controls the number of bytes used to identify tags, so if you change
it the existing data becomes unusable.  Conceivably, we could change the
encoding to say how many bits are being used, but that wastes a lot of bits
in HBase on things that are largely static.  Another option which might
waste fewer bits is to have some schema version id in the row key which
then was looked up in a table to get the sizes.  However, I'm not sold on
this being all that useful to support changing it without modifying old
data.

-- 
John A. Tamplin

oozie:
I agree that for the reasons @jtamplin pointed out, `MAX_NUM_TAGS` should not be configurable. It would create a world of supportability problems.

[...]

Issue #450, opened 2015-02-26
Tracking Invalid Values


DawgTool:
Hi All,

Trying to track invalid and illegal arguments, however I am not having any success.
I currently have the root logger set in debug, along with hb,zk,as.  This doesn't seem to give me anything in regards to catching these errors.  Any suggestions?

tsd.rpc.errors 1424987992 1302 type=invalid_values host=localhost
tsd.rpc.errors 1424987992 2268 type=illegal_arguments host=localhost

manolama:
How are you trying to write data, over HTTP or Telnet? Right now TSDB only returns errors like those to the client and doesn't log them locally (though it does increment the proper error counters). If you use Telnet you can read the errors off the socket (though you can't really associate them with a data point) or if you use HTTP it will return them if you ask for details. Of you do want to log them, I can point you at the locations where you can add some log lines.

manolama:
Closing post @oozie's merge.

PR #467, opened 2015-03-22
Log illegal put arguments at DEBUG level.

oozie:
The TSD already returns an illegal argument message to the client, 
```
$ nc number5 4242
put mymetric 1427014820 1 host=thursday a=a b=b c=c d=d e=e f=f g=g i=i
put: illegal argument: Too many tags: 9 maximum allowed: 8, tags: {f=f, g=g, d=d, e=e, host=thursday, b=b, c=c, a=a, i=i}
```
but when a collector framework does not surface the problem, it can be hard to realize otherwise, as reported in #437. Adding the error message at a DEBUG level to make troubleshooting of missing metrics easier.
```
02:34:12.886 DEBUG [PutDataPointRpc.execute] - put: illegal argument: Too many tags: 9 maximum allowed: 8, tags: {f=f, g=g, d=d, e=e, host=thursday, b=b, c=c, a=a, i=i}
```

### DIFF
--- a/src/core/IncomingDataPoints.java
+++ b/src/core/IncomingDataPoints.java
@@ -93,7 +93,7 @@ final class IncomingDataPoints implements WritableDataPoints {
    */
   static void checkMetricAndTags(final String metric, final Map<String, String> tags) {
     if (tags.size() <= 0) {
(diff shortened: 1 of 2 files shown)


oozie:
This will also help address #450

PR #468, opened 2015-03-25
Two tiny fixes for 'tsdb import' and stats reporting on UID's > 3 bytes.

hynd:


manolama:
Merged into next and put with fa8edf7226d59521d8ac2354a05b3c5927fd92c1

PR #472, opened 2015-03-29
Fix #452 Protect the UI from getting stuck when no tags are returned.

### LINKED ISSUE #452: GUI stops responding after a while.
After a long time running the web interface GUI seems to stop responding to changes in the plotting parameters.  I can change things in the URL query parameters (the GET parameters) and then reload the browser, and this works.  Everything changed via the GUI, however, does not respond.  What's more weird, is that I have other tabs open that don't seem to be staled out at all.  Copying the staled URL and pasting it into a fresh tab has no effect either.  What does seem to work is starting over from a fresh http://host:4242 and then using the GUI to build up the time series that I want to plot.  This, however, is time consuming.

Has anyone else experienced this and found a way to fix it when it happens?

Thanks!

oozie:
When one of the queries returns no datapoints, the suggested tags for it are empty and therefore not set (they remain null). QueryUI code assumes that they will always be at least a 0-element array, which is not true.

1. To reproduce, take a standard tsd metric and graph it.
2. Clone this metric and under host tag enter "eth0", which is a legit tag-value pair, but will likely return 0 data points.
3. Try to change the time range on the ui - the UI should be stuck at this point.

An example bad query: http://j.mp/19f04EE - just replace the host:port with your TSD's address.

With this change, the UI isn't getting stuck in such situations.

### DIFF
--- a/src/tsd/client/QueryUi.java
+++ b/src/tsd/client/QueryUi.java
@@ -956,9 +956,11 @@ public class QueryUi implements EntryPoint, HistoryListener {
             }
             final MetricForm metric = (MetricForm) widget;
             final JSONArray tags = etags.get(i).isArray();
-            final int ntags = tags.size();
-            for (int j = 0; j < ntags; j++) {
-              metric.autoSuggestTag(tags.get(j).isString().stringValue());
+            // Skip if no tags were associated with the query.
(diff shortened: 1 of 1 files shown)


oozie:
Fixes #452

PR #479, opened 2015-04-02
Fix startup on boot via init script (RPM)

hynd:
NAME=`basename $0` only works if you're starting opentsdb by hand.
On boot `basename $0` actually ends up being "S80opentsdb" (since it's run from the /etc/rcX.d symlinks rather than /etc/init.d) - the init script can't find /etc/opentsdb/S80opentsdb.conf, writes its logs to S80opentsdb-xxx etc.

Instead we can default to a static 'opentsdb' for NAME, and let it be overridden in a /etc/sysconfig/`basename $0` if you want to run multiple instances.

### DIFF
--- a/build-aux/rpm/init.d/opentsdb
+++ b/build-aux/rpm/init.d/opentsdb
@@ -28,18 +28,14 @@
 # Source init functions
 . /etc/init.d/functions
 
-# Set this so that you can run as many opentsdb instances you want as long as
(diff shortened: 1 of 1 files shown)


hynd:
I don't believe so - I didn't see the `basename $0` bits in the DEB init scripts and it usually starts fine on boot on my Debian machines.

Issue #473, opened 2015-03-30
Double-precision floating-point values?


lburg:
According to the [documentation](http://opentsdb.net/docs/build/html/user_guide/writing.html#floating-point-values) only single-precision floating-points are supported, double-precision will come in a future release.

However, looking at the OpenTSDB core documentation, I see that there is a [addPoint](http://opentsdb.net/docs/javadoc/net/opentsdb/core/TSDB.html#addPoint%28java.lang.String, long, double, java.util.Map%29) method to add a double-precision floating-point value to the database. The PR #380 also refers to this by patching a comment. This PR was submitted in August 2014 and the `addPoint` method was added in version 1.2. I'm using version 2.0 and adding a double-precision floating-point value through telnet stores only a single-precision value (I'm running HBase 1.0).

Do you have an estimate release date for double-precision support?

manolama:
Just need to review the PR and we can likely get it into 2.2 where I'm aiming for a May RC1 release.

IDerr:
You can close this @manolama

IDerr:
Hi @lburg,
if your problem has been resolved, could you please close this issue ?

Thanks :)

lburg:
Whoops, sorry! Closing it, thanks!

PR #483, opened 2015-04-08
Handle double precision floating point via the API -- refs #473

lburg:
Add a `fitsInFloat` function similar to `looksLikeInteger` to determine whether a floating point value should be stored in a Float or a Double. Add a few tests and clean up some trailing whitespaces.

There is also a call to `looksLikeInteger` in [src/tools/TextImport.java](https://github.com/OpenTSDB/opentsdb/blob/master/src/tools/TextImporter.java#L151), but I'm not sure what to do here. It doesn't seem like there is a way to add doubles in the `WritableDataPoints` interface.

It seemed like a pretty easy change so I'm wondering: is there a reason this wasn't implemented sooner? Am I missing something?

### DIFF
--- a/src/core/Tags.java
+++ b/src/core/Tags.java
@@ -295,7 +295,7 @@ public final class Tags {
       System.arraycopy(row, pos + name_width, tmp_value, 0, value_width);
       deferreds.add(tsdb.tag_values.getNameAsync(tmp_value));
     }
-    
+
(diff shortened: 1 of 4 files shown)


lburg:
Closing this as I realized I changed the `master` branch instead of the `next` branch. See #485

PR #485, opened 2015-04-09
Handle double precision floating point via the API -- refs #473

lburg:
Add a `fitsInFloat` function similar to `looksLikeInteger` to determine whether a floating point value should be stored in a Float or a Double. Add a few tests.

There is also a call to `looksLikeInteger` in [src/tools/TextImport.java](https://github.com/OpenTSDB/opentsdb/blob/master/src/tools/TextImporter.java#L151), but I'm not sure what to do here. It doesn't seem like there is a way to add doubles in the `WritableDataPoints` interface.

It seemed like a pretty easy change so I'm wondering: is there a reason this wasn't implemented sooner? Am I missing something?

### DIFF
--- a/src/core/Tags.java
+++ b/src/core/Tags.java
@@ -637,4 +637,19 @@ public final class Tags {
     return true;
   }
 
+  /**
+   * Returns true if the given string can fit into a float.
+   * @param value The String holding the float value.
+   * @return true if the value can fit into a float, false otherwise.
+   * @throws NumberFormatException if the value is not numeric.
+   * @since 2.0.2
+   */
+  public static boolean fitsInFloat(final String value) {
+    final float f = Float.parseFloat(value);
+    final String converted = Float.toString(f);
+
+    // this will be false if there was a loss of precision.
+    return value.equals(converted);
+  }
+
 }
--- a/src/tsd/PutDataPointRpc.java
+++ b/src/tsd/PutDataPointRpc.java
@@ -145,9 +145,12 @@ final class PutDataPointRpc implements TelnetRpc, HttpRpc {
         if (Tags.looksLikeInteger(dp.getValue())) {
           tsdb.addPoint(dp.getMetric(), dp.getTimestamp(), 
               Tags.parseLong(dp.getValue()), dp.getTags());
-        } else {
+        } else if (Tags.fitsInFloat(dp.getValue())) {
           tsdb.addPoint(dp.getMetric(), dp.getTimestamp(), 
               Float.parseFloat(dp.getValue()), dp.getTags());
+        } else {
+          tsdb.addPoint(dp.getMetric(), dp.getTimestamp(),
+              Double.parseDouble(dp.getValue()), dp.getTags());
         }
         success++;
       } catch (NumberFormatException x) {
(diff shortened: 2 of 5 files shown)


src/core/Tags.java
```diff
@@ -637,4 +637,19 @@ public static boolean looksLikeInteger(final String value) {
...
+   * @return true if the value can fit into a float, false otherwise.
+   * @throws NumberFormatException if the value is not numeric.
+   * @since 2.0.2
+   */
+  public static boolean fitsInFloat(final String value) {
+    final float f = Float.parseFloat(value);
+    final String converted = Float.toString(f);
+
+    // this will be false if there was a loss of precision.
+    return value.equals(converted);
```
manolama: I don't think this is the proper way to determine if the precision is lost as some floating point values won't survive the change due to the fact that some values can't be represented in base 2. Instead I saw one way was to parse to Double, then cast to float and see if the result was infinite or not. But that didn't seem to help with negative floats so we'd have to figure that one out.

http://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html

manolama:
One comment on the detection method then we'd need some tests in TestTSDB to make sure the values can be written then read back out and some tests in TestTsdbQuery for the same.
Thanks!

src/core/Tags.java
lburg: How about reading straight to double, then taking the absolute value of this number and check it against `Float.MAX_VALUE` ? If it's greater then we keep it as double, otherwise we cast it down to float.

jtamplin: So there are two ways that doubles represent more values - the exponent range is larger, and it has more mantissa bits.  The exponent range is easy enough, and @manolama 's suggestion of casting and comparing to infinity will catch that (not sure what the issue is with negative floats -  a value less than Float.MIN_VALUE should be converted to Float.NEGATIVE_INFINITY; if this really is an issue, then you can look at Double.doubleToLongBits and extract the exponent yourself).  The only way to be certain you aren't losing any precision is if the extra mantissa bits are all 0, but in practice representing decimal values that is not ever going to be the case, because most decimal fractions are infinitely long repeating binary fractions.  Instead, what you want is some acceptable level of precision loss, and it probably is an absolute value (ie, if the difference is 1e-10 you probably don't care, while if it is 1e10 you probably do, even if the number of bits in the mantissa being lost is the same.  So my suggestion for the test would be casting to float and check if the absolute value difference is greater than .0078125 (2^-7), but this really should be configurable since it depends on what data I am storing (and actually should be per-metric, but that is probably unacceptable overhead).

lburg:
I added a few tests to `TestTSDB.java` (mostly a copy/paste of the floating point tests, using doubles instead of floats), however something seemed weird to me. `assertEquals` for floating points values only accept doubles as input, doesn't that make the tests using floats a bit unstable?

lburg:
Also, regarding the tests in `TestTsdbQuery`, should I write a double variant of every float tests? Like `runMimMaxFloat`, `runMimMaxFloatOffset`, etc.? I'll write a few functions similar to `storeFloatTimeSeriesSeconds` using doubles for now.

johann8384:
@manolama Is this ready to be merged?

manolama:
Cleaned up and merged in a0d8049607584f35446fbc171e86b32e31a901cf. Thanks!

PR #493, opened 2015-04-16
Support CSV data type in in TSDB Response (header Accept: text/csv)

synle:
This is a sample:
Header
Acccept: text/csv

screen shot 2015-04-16 at 4 44 37 pm

I am working on unit tests, will update once everything is done.


synle:
Serializer contains 2 main methods: parse (parsing input), and format (formatting responses).

Supporting CSV getting complicated since we need to support the parse and format function of the csv format as well. These are not really hard to do, but we might need to come up with format for both parse and format functions. 

Shall we stick with JSON for parsing and only provide CSV format for formatQueryAsyncV1.

mxk1235:
this is a pretty cool feature

Issue #81, opened 2011-10-14
Deleting data from HTTP / TCP


matteoredaelli:
is it possible to delete data via HTTP / TCP requests?

theoryno3:
Would you guys be able to point me to the portion of the code (or perhaps the documentation) regarding this feature? This feature request hasn't been closed, so my initial assumption (admittedly, without much digging) is that this feature has not yet been implemented. Thanks!

IDerr:
This one is resolved @manolama
 You can close this :D

IDerr:
You can the delete argument on call 
You can check the documentation.
http://opentsdb.net/docs/build/html/api_http/query/index.html
@manolama you can close this

PR #496, opened 2015-04-20
Add possibility to DELETE datapoints through the query endpoint

lburg:
Give the user a way to delete datapoints though the `api/query` endpoint  using a `DELETE` HTTP request. It's basically a copy/paste/trim from the [doDump](https://github.com/OpenTSDB/opentsdb/blob/7a893b165f5aba66a2040deb8ec34d3c9eca7a32/src/tools/DumpSeries.java#L88) function, as I couldn't find a way to do it otherwise (if there is even another way?).

A few problems with this PR:

- No tests, I couldn't figure out how to add some sample data points in the database. I looked around `TestQueryRpc.java` but there doesn't seem to be any tests that do a `GET` request. How should I proceed?
- The `q.setEndTime(Integer.MAX_VALUE);` line. Our use case was to remove every data points after a given timestamp. However if we do not specify an end timestamp, OpenTSDB uses the current server time. As we can have data in the future (estimations about energy consumption in our case), I manually set the end timestamp so it would delete everything. I believe this should not be the default use case, maybe add a URL parameter like `drop=True` which drops every datapoints?
- Not enough log info?
- Too big of a try-catch block surrounding the scan? I saw several places in the code that just wrapped exception-raising code with a try-catch and just printed `"Shouldn't be here"` so I did the same.

I didn't return after the delete operations so it would return an empty set, confirming the deletion.

### DIFF
--- a/src/tsd/QueryRpc.java
+++ b/src/tsd/QueryRpc.java
@@ -21,6 +21,10 @@ import java.util.List;
 import java.util.Map;
 
 import org.hbase.async.Bytes.ByteMap;
+import org.hbase.async.DeleteRequest;
+import org.hbase.async.HBaseClient;
+import org.hbase.async.KeyValue;
+import org.hbase.async.Scanner;
 import org.jboss.netty.handler.codec.http.HttpMethod;
 import org.jboss.netty.handler.codec.http.HttpResponseStatus;
 import org.slf4j.Logger;
@@ -32,6 +36,7 @@ import com.stumbleupon.async.DeferredGroupException;
 
 import net.opentsdb.core.DataPoints;
 import net.opentsdb.core.IncomingDataPoint;
+import net.opentsdb.core.Internal;
 import net.opentsdb.core.Query;
 import net.opentsdb.core.RateOptions;
 import net.opentsdb.core.TSDB;
@@ -43,6 +48,7 @@ import net.opentsdb.meta.TSMeta;
 import net.opentsdb.meta.TSUIDQuery;
 import net.opentsdb.uid.NoSuchUniqueName;
 import net.opentsdb.uid.UniqueId;
+import net.opentsdb.uid.UniqueId.UniqueIdType;
 import net.opentsdb.utils.JSON;
 
 /**
@@ -68,8 +74,9 @@ final class QueryRpc implements HttpRpc {
   public void execute(final TSDB tsdb, final HttpQuery query) 
     throws IOException {
     
-    // only accept GET/POST
-    if (query.method() != HttpMethod.GET && query.method() != HttpMethod.POST) {
+    // only accept GET/POST/DELETE
+    if (query.method() != HttpMethod.GET && query.method() != HttpMethod.POST &&
+        query.method() != HttpMethod.DELETE) {
       throw new BadRequestException(HttpResponseStatus.METHOD_NOT_ALLOWED, 
           "Method not allowed", "The HTTP method [" + query.method().getName() +
           "] is not permitted for this endpoint");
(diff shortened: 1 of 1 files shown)


lburg:
Refs #81

src/tsd/QueryRpc.java
```diff
@@ -115,14 +122,45 @@ private void handleQuery(final TSDB tsdb, final HttpQuery query) {
...
       tsdbqueries = data_query.buildQueries(tsdb);
     } catch(NoSuchUniqueName ex) {
       throw new BadRequestException(ex);
     }
+
     final int nqueries = tsdbqueries.length;
+    if (query.method() == HttpMethod.DELETE) {
+      final ArrayList<Deferred<Object>> deferredsDelete =
+        new ArrayList<Deferred<Object>>(nqueries);
+      final byte[] table = tsdb.getConfig().getString("tsd.storage.hbase.data_table").getBytes();
```
manolama: Use ``tsdb.dataTable()`` instead.

src/tsd/QueryRpc.java
```diff
@@ -115,14 +122,45 @@ private void handleQuery(final TSDB tsdb, final HttpQuery query) {
...
+
     Query[] tsdbqueries;
     try {
       tsdbqueries = data_query.buildQueries(tsdb);
     } catch(NoSuchUniqueName ex) {
       throw new BadRequestException(ex);
     }
+
     final int nqueries = tsdbqueries.length;
+    if (query.method() == HttpMethod.DELETE) {
```
manolama: I'd prefer if you break this out into a separate function please.
Also, add a setting in the config class, something like "tsd.http.query.allow_delete" that is set to false by default. We don't want folks deleting on accident :)

src/tsd/QueryRpc.java
```diff
@@ -115,14 +122,45 @@ private void handleQuery(final TSDB tsdb, final HttpQuery query) {
...
+      final ArrayList<Deferred<Object>> deferredsDelete =
+        new ArrayList<Deferred<Object>>(nqueries);
+      final byte[] table = tsdb.getConfig().getString("tsd.storage.hbase.data_table").getBytes();
+      final HBaseClient client = tsdb.getClient();
+
+      LOG.info("deleting from table=" + Arrays.toString(table));
+      try {
+        for (final Query q : tsdbqueries) {
+          q.setEndTime(Integer.MAX_VALUE);
+          final Scanner scanner = Internal.getScanner(q);
```
manolama: Instead of scanning HBase twice, you should pass a delete flag to the scanners in TsdbQuery and SaltScanner (Put branch) to perform the scan asynchronously and once instead of twice. This will block the thread every time we wait for HBase to fetch the next chunk of data, reducing the # of queries we can handle per second.

manolama:
Thanks for the work on this! Regarding your bullet points,

* Take a look at MockBase and the unit tests for the Fsck utility. That will let you write and delete dummy data.
* It's actually a bigger problem with the default scanner implementation in that we'll delete some data before and after what the user explicitly asks for since we're deleting entire rows instead of columns. Ideally we would make sure to only delete columns within the specified time range but that's a fair bit of work and can be left as a TODO.
For now though, we don't want to set the end time to max for everyone so you could leave that is. For your case you could probably use set your query end time to 9999999999
* The logging is ok, but you may want to set the rowkey log to debug as it could delete tons of rows. At the end you should tally the rows deleted and maybe return that to the caller

The only other comments I had were adding a config flag to make sure people don't delete on accident, make sure it's async all the way through (I'm trying to get all of the HTTP calls async) and base off the "put" branch as that will be where this will go. Thanks!

src/tsd/QueryRpc.java
lburg: So I should modify the `ScannerCB` class in both SaltScanner and TsdbQuery to delete the data if a flag was given, and then return an empty `Span` for each row? If so, I don't see how to integrate the `DeleteRequest` in this process. Any advice?

lburg:
I'm closing this PR since it's based on the wrong branch, see #503

Issue #498, opened 2015-04-21
Missing info from /api/uid/tsmeta when using the metric query style


lburg:
I'm using a version of OpenTSDB compiled straight from the `next` branch.

I have a metric named `toto` in which I added a few points through telnet. I activated the automatic creation of time series meta data objects and I want to retrieve the infos after my insertions. I issued a `GET` request using `curl`:

    curl -X GET "http://62.210.69.193:4242/api/uid/tsmeta?m=toto"

Here's what I got:

```json
[  
   {  
      "tsuid":"050BCD00000E000014",
      "metric":{  
         "uid":"050BCD",
         "type":"METRIC",
         "name":"toto",
         "description":"",
         "notes":"",
         "created":0,
         "custom":null,
         "displayName":""
      },
      "tags":[  
         {  
            "uid":"00000E",
            "type":"TAGK",
            "name":"v",
            "description":"",
            "notes":"",
            "created":1429280021,
            "custom":null,
            "displayName":""
         },
         {  
            "uid":"000014",
            "type":"TAGV",
            "name":"1",
            "description":"",
            "notes":"",
            "created":1429280021,
            "custom":null,
            "displayName":""
         }
      ],
      "description":"",
      "notes":"",
      "created":1429280022,
      "units":"",
      "retention":0,
      "max":"NaN",
      "min":"NaN",
      "displayName":"",
      "dataType":"",
      "lastReceived":0,
      "totalDatapoints":0
   }
]
```

As we can see, the `lastReceived` and `totalDatapoints` keys are not filled with the correct values, which I thought was weird. I then issued a second request, this time using the TSUID style:

    curl -X GET "http://62.210.69.193:4242/api/uid/tsmeta?tsuid=050BCD00000E000014"

And this was the answer:

```json
{  
   "tsuid":"050BCD00000E000014",
   "metric":{  
      "uid":"050BCD",
      "type":"METRIC",
      "name":"toto",
      "description":"",
      "notes":"",
      "created":0,
      "custom":null,
      "displayName":""
   },
   "tags":[  
      {  
         "uid":"00000E",
         "type":"TAGK",
         "name":"v",
         "description":"",
         "notes":"",
         "created":1429280021,
         "custom":null,
         "displayName":""
      },
      {  
         "uid":"000014",
         "type":"TAGV",
         "name":"1",
         "description":"",
         "notes":"",
         "created":1429280021,
         "custom":null,
         "displayName":""
      }
   ],
   "description":"",
[...]

PR #499, opened 2015-04-21
Fix missing info from /api/uid/tsmeta - refs #498

lburg:
Unfortunately this breaks the [parseFromColumnWithUIDMetaNSU](https://github.com/OpenTSDB/opentsdb/blob/next/test/meta/TestTSMeta.java#L400) test as the `NoSuchUniqueId` exception is not raised anymore since `LoadUIDs.call` [returns early](https://github.com/OpenTSDB/opentsdb/blob/next/src/meta/TSMeta.java#L837) due to `meta` being `null`. I didn't know how to fix the test, whether to keep it as-is and try to raise the same exception again, or change it to make the tsuid `000001000001000002` a valid TSUID missing a UID meta object.

What's the best approach?

### DIFF
--- a/src/meta/TSMeta.java
+++ b/src/meta/TSMeta.java
@@ -406,20 +406,20 @@ public final class TSMeta {
       throw new IllegalArgumentException("Empty column value");
     }
 
-    final TSMeta parsedMeta = JSON.parseToObject(column.value(), TSMeta.class);
+    final TSMeta parsed_meta = JSON.parseToObject(column.value(), TSMeta.class);
     
     // fix in case the tsuid is missing
-    if (parsedMeta.tsuid == null || parsedMeta.tsuid.isEmpty()) {
-      parsedMeta.tsuid = UniqueId.uidToString(column.key());
+    if (parsed_meta.tsuid == null || parsed_meta.tsuid.isEmpty()) {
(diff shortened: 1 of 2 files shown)


manolama:
Nice catch, I'd follow the second approach for the unit test and make sure it throws the NSUI.

src/meta/TSMeta.java
```diff
@@ -405,24 +405,21 @@ public ValidateCB(final TSMeta local_meta) {
     if (column.value() == null || column.value().length < 1) {
       throw new IllegalArgumentException("Empty column value");
     }
-    
-    final TSMeta meta = JSON.parseToObject(column.value(), TSMeta.class);
+
+    final TSMeta parsedMeta = JSON.parseToObject(column.value(), TSMeta.class);
```
manolama: Minor nitpick, ``parsed_meta`` :)

lburg: Oops, I've been tricked by the methods' names!

lburg:
After reading the `before` method more carefully, I managed to fix the test. Let me know if it was done correctly!

manolama:
Merged in b1c32620cd008572fcf9e7ed8c9701c22eac85b2

PR #503, opened 2015-04-23
Async delete of datapoints through HTTP

lburg:
Follow up of #496, based on the put branch.

Still missing tests, I will get to that next.


gucki:
Will this be merged anytime soon? I'm using opentsdb in a cloud environment and I have a data of many hosts which have been deleted long time ago. It be really nice if ther'd be an easy way of deleting all that obsolete data. Like deleting all data (any metric) with tag host = xxx with a single query.

lburg:
Well one of the test is broken so I should get to fix that before merging. Other than that, do you see anything missing @manolama ?

manolama:
Cleaned up and merged in 7180bb5c5410ebf26a9f903ef1a6f4d5897a25da

lburg:
Great, thanks!

PR #518, opened 2015-06-05
Updates 'search/lookup' to obey limit

lexh:
This commit updates the ['search/lookup' HTTP API endpoint](http://opentsdb.net/docs/build/html/api_http/search/lookup.html) to obey the SearchQuery limit property. Now, a request can contain an optional 'limit' parameter which sets an upper bound on the number of results returned.

The default value remains 25, but is now actually enforced.

Context: https://github.com/grafana/grafana/issues/1250#issuecomment-109185678


manolama:
Thanks a ton for this! Merged in d5f79860b16fddfb11ad67a0c54cf5dbe6663d2f

PR #529, opened 2015-06-21
Allow config file to determine size of UID's for metrics, tag-keys an…

aabes:
…d tag-values

### DIFF
--- a/src/core/TSDB.java
+++ b/src/core/TSDB.java
@@ -65,11 +65,13 @@ public final class TSDB {
   /** Charset used to convert Strings to byte arrays and back. */
   private static final Charset CHARSET = Charset.forName("ISO-8859-1");
   private static final String METRICS_QUAL = "metrics";
-  private static final short METRICS_WIDTH = 3;
   private static final String TAG_NAME_QUAL = "tagk";
-  private static final short TAG_NAME_WIDTH = 3;
   private static final String TAG_VALUE_QUAL = "tagv";
-  private static final short TAG_VALUE_WIDTH = 3;
+
+  private static short uid_metrics_width;
+  private static short uid_tagk_width;
+  private static short uid_tagv_width;
+
 
   /** Client for the HBase cluster to use.  */
   final HBaseClient client;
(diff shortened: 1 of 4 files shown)


aabes:
seems to be more intricate that I first hoped - since METRIC_WIDHT and its friends are assumed to be static values of the TSDB class in quite a few places.

I can convert the various places that make this assumption, to leverage the config class....
But is this general direction acceptable?

jtamplin:
Personally if we are going to do something like this, I would prefer to use
a variable length encoding scheme instead, and then make that the default
for new installs so things don't just fall over at some threshold.  For
example, take two bits in the first byte to encode 2-5 byte ids.

John A. Tamplin (phone)

aabes:
@jtamplin - the UID size appears to be pivotal to rowkey structure.
if I correctly understand your approach, each row effectively could use a different encoding?
If that's the general thrust, it might be very problematic in terms of performance, and impact on the overall design of the HBase schema, here's why I think so: 
There are multiple places where assumptions are made about the structure of rowkeys, e.g.:
  https://github.com/CloudHealth/opentsdb/blob/uid-from-config/src/core/TsdbQuery.java#L540 and
https://github.com/CloudHealth/opentsdb/blob/uid-from-config/src/core/TsdbQuery.java#L847

these assumptions allow for limiting the set of rows being scanned, and efficiently compared. Having to deal with different row keys being structured differently (i.e. variably encoded UID's) would require rethinking these optimizations.

The change I was proposing was much less invasive into the overall design.

aabes:
The current code and tests make the assumption that TSDB is statically capable of providing the UID sizes... the changes are much more pervasive than I initially thought (though don't modify the overall design). 
Closing PR until have fixed all the places that make above mentioned assumption....

jtamplin:
On Mon, Jun 22, 2015 at 3:01 PM, aabes <notifications@github.com> wrote:

> @jtamplin <https://github.com/jtamplin> - the UID size appears to be
> pivotal to rowkey structure.
> if I correctly understand your approach, each row effectively could use a
> different encoding?
> If that's the general thrust, it might be very problematic in terms of
> performance, and impact on the overall design of the HBase schema, here's
> why I think so:
> There are multiple places where assumptions are made about the structure
> of rowkeys, e.g.:
>
> https://github.com/CloudHealth/opentsdb/blob/uid-from-config/src/core/TsdbQuery.java#L540
> and
>
> https://github.com/CloudHealth/opentsdb/blob/uid-from-config/src/core/TsdbQuery.java#L847
>
> these assumptions allow for limiting the set of rows being scanned, and
> efficiently compared. Having to deal with different row keys being
> structured differently (i.e. variably encoded UID's) would require
> rethinking these optimizations.
>
These assumptions aren't what allow limiting the rows being scanned - what
does that is building a start and an end key.  Since the encoding of a
particular UID is deterministic, you can still do that.  Certainly, code
would need to change.  In your first example, you simply have to know what
metric UID you are encoding before you allocate the space for the row key
prefix and allocate accordingly.  The second one is even easier, just look
at the size bits from the first byte to see how many more bytes to
compare.  Where you would have a slight performance impact is that the
timestamp and tagk/tagv values aren't at fixed locations anymore, but
scanning a few bytes in RAM is negligible to the cost of going to HBase
anyway.

We could even make all size bits zero mean that it is 3-byte UIDs, so any
existing data stores that didn't use too many UIDs would be able to keep
all existing data without change.  That is probably more likely if we only
used a single bit, so it could be like 0=3-byte UIDs, 1=6-byte UIDs.  If we
used two bits, it could be 00=3, 01=4, 10=5, 11=6, but anyone with more
than 2M UIDs would have to rewrite or discard their data.

> The change I was proposing was much less invasive into the overall design.
>
Sure, but doesn't really buy you anything other than not having to
recompile to change the value.  I was proposing to eliminate the problem of
having to chose entirely.  Nobody gets to a cliff where things will stop
functioning when they hit it, and then they either have to rewrite all
their old data or discard it when they make the change.

-- 
John A. Tamplin

PR #537, opened 2015-07-13
Fix RPM init script. 

noharm:
Invert user comparision. Chown log & lock file when user is not root


manolama:
But don't you have to be root to chown something? Usually?

noharm:
Yes you do. Only root can chown files belonging to other users. It's the basis of unix filesystem security.


Verzonden vanaf Samsung Mobile.


-------- Oorspronkelijk bericht --------
Van: Chris Larsen
Datum:05-09-2015 00:40 (GMT+01:00)
Aan: OpenTSDB/opentsdb
Cc: Harm Kroon
Onderwerp: Re: [opentsdb] Fix RPM init script. (#537)


But don't you have to be root to chown something? Usually?

-

noharm:
... Sorry for editing / deleting posts. But I had to retrace my thoughts why I made / proposed that change.
I was using this under the assumption that init scripts are allways executed as root. But after going through the code again, I have to conclude this is intended to also be executed by a non-privileged account to running the TSD. I'll test again why I needed this for.

PR #548, opened 2015-08-17
Added Downsample normalize on week month and year

moarcaccio:
downsample normalize on week, month and year on non-UTC times using Java Calendar.
It needs additional refinements, but it works, the resulting timestamp will be rounded off at the beginning of the week (7days), month (31days) or year (365days).

### DIFF
--- a/src/core/Downsampler.java
+++ b/src/core/Downsampler.java
@@ -13,6 +13,7 @@
 package net.opentsdb.core;
 
 import java.util.NoSuchElementException;
+import java.util.Calendar;
 
 
 /**
--- a/src/core/Downsampler.java
+++ b/src/core/Downsampler.java
@@ -192,8 +192,55 @@ public class Downsampler implements SeekableView, DataPoint {
     }
 
     /** Returns timestamp aligned by interval. */
-    private long alignTimestamp(long timestamp) {
-      return timestamp - (timestamp % interval_ms);
+     private long alignTimestamp(long timestamp) {
+       // 1 year (365days), align the timestamp at the first day of year.
+       if (interval_ms == 31536000000L)
+         return toFirstDayOfYear(timestamp);
+	   // 1 month (31days), align the timestamp at the first day of month. 
+       else if (interval_ms == 2678400000L)
+         return toFirstDayOfMonth(timestamp);
+	   // 1 week (7days), align the timestamp at the first day of week.
+	   else if (interval_ms == 604800000L)
+		 return toFirstDayOfWeek(timestamp);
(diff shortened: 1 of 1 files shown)


manolama:
Cool, thanks for taking a stab at this! Could you do a couple of things:
* Point it at the Next branch please and work off that?
* Add a configuration flag like "tsd.query.downsample.use_calendar" that would enable this feature or use the current default for backwards compatibility? Some folks may depend on the current behavior.
* Add a per-query flag that would override the above
* Add daily support (some timezones are off by 30 minutes, etc [including North Korea now])
* Sign the contributors agreement

src/core/Downsampler.java
```diff
@@ -192,8 +193,55 @@ private long getIntervalTimestamp() {
...
+	   // 1 week (7days), align the timestamp at the first day of week.
+	   else if (interval_ms == 604800000L)
+		 return toFirstDayOfWeek(timestamp);
+	   // default timestamp normalization (tsdb v2.1.0)
+       else
+         return timestamp - (timestamp % interval_ms);
+    }
+    
+    /** Returns timestamp aligned at the first day of month. */    
+    private long toFirstDayOfMonth(long timestamp) {
```
manolama: Move these to the DateTime class as we may want to share them

src/core/Downsampler.java
```diff
@@ -192,8 +193,55 @@ private long getIntervalTimestamp() {
...
+		Calendar c = Calendar.getInstance();
```
manolama: Be sure to pass in the time zone if set.

src/core/Downsampler.java
```diff
@@ -192,8 +193,55 @@ private long getIntervalTimestamp() {
     }
 
     /** Returns timestamp aligned by interval. */
-    private long alignTimestamp(long timestamp) {
-      return timestamp - (timestamp % interval_ms);
+     private long alignTimestamp(long timestamp) {
+       // 1 year (365days), align the timestamp at the first day of year.
+       if (interval_ms == 31536000000L)
```
manolama: It should handle multiples as well, e.g. ``2y`` or ``4w``

mxk1235:
this is super useful.  is this going to make it into 2.2?

cpdevoto:
This pull request is incomplete.  I have issued a new pull request (#657) which builds on the work initiated by @moarcaccio, but which also addresses all of the enhancements requested by @manolama, and resolves a few defects with the original pull request.

manolama:
Merged @cpdevoto's in 5012f7da15bb872bae3c4fd6129b10d38c7837c4

PR #562, opened 2015-09-09
add first and last aggregator

andyflury:
It looks like first and last aggregators are currently missing. This is important when trying to aggregate financial tick data into high-low-open-close Bars. Would you be so kind as to verify these changes make sense? Thanks Andy Flury

### DIFF
--- a/src/core/Aggregators.java
+++ b/src/core/Aggregators.java
@@ -65,6 +65,14 @@ public final class Aggregators {
    * if timestamps don't line up instead of interpolating. */
   public static final Aggregator MIMMAX = new Max(
       Interpolation.MIN, "mimmax");
+
+  /** Aggregator that returns the first data point. */
+  public static final Aggregator FIRST = new First(
+      Interpolation.ZIM, "first");
+
+  /** Aggregator that returns the first data point. */
+  public static final Aggregator LAST = new Last(
+      Interpolation.ZIM, "last");
   
   /** Maps an aggregator name to its instance. */
   private static final HashMap<String, Aggregator> aggregators;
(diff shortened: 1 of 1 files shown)


manolama:
Yeah, I think they make sense if you're using them on downsamplers. But it may be useful to throw an exception if someone tries to use them for group by aggregation as it doesn't work for grouping as we don't order the results in any particular fashion.
Also would you mind throwing in a couple of UTs please? Thanks!

andyflury:
Hi Chris

 

Sure, I’ll be glad to add this but how do I know if the Aggregation method is getting called by grouping instead of down sampling?

 

Also, would you be so kind as to point me to some similar unit tests that I could reuse to add additional tests?

 

Regards

Andy

manolama:
Thanks Andy! Check out https://github.com/OpenTSDB/opentsdb/blob/next/test/core/TestTsdbQueryDownsample.java for unit tests around downsampling, just plugin your first/last aggs.

For checking to see if the aggregator would be used for group bys, take a look at https://github.com/OpenTSDB/opentsdb/blob/next/src/core/TsdbQuery.java#L300 and the ``TsdbQuery.setTimeseries()`` methods.

andyflury:
Thanks, I had a look at the links but I’m afraid I’ll have to leave that up to you guys. Aggregators.java was quite straight forward to modify but it looks like modifying the UnitTests and TsdbQuery would require a bit more knowledge of the internals of TSDB.

manolama:
Added in 535b7e553812d1f2c13377d57e352e5c131a4097. Thanks!

PR #565, opened 2015-09-10
Allow overriding the maximum number of tags via config

hynd:
Continues to default to 8.

Since precedent has been set for making footgun configs with UID widths and salting, having this configurable would also avoid some custom patching :)

Couple of questions...
* Is the config name (`tsd.storage.max_tags`) OK?
* Would you prefer me to switch any other code using Const.MAX_NUM_TAGS to the Const.MAX_NUM_TAGS() i added for the tests?

Cheers,
-kieren

### DIFF
--- a/src/core/Const.java
+++ b/src/core/Const.java
@@ -19,8 +19,26 @@ public final class Const {
   public static final short TIMESTAMP_BYTES = 4;
 
   /** Maximum number of tags allowed per data point.  */
-  public static final short MAX_NUM_TAGS = 8;
-  // 8 is an aggressive limit on purpose.  Can always be increased later.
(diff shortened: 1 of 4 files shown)


manolama:
Ah yeah, missed that one, thanks!
I don't think we ever had any tests around the max tags. If you grep for the old MAX_NUM_TAGS and find it anywhere, please tweak it :)

hynd:
Doned! Only found it in one place.

hynd:
Hi @manolama - any tips on how i can fix up the tests? (functionally the code works as expected).  Thanks!

manolama:
Fixed up the UTs and merged in ad776f61cb44ab3f01b41aa7278d78583b46c3f9. Thanks!

Issue #568, opened 2015-09-16
tsd.storage.salt.width > 0 breaks /api/search/lookup when tsd.core.meta.enable_realtime_ts is true


Dieken:
I use latest "next" branch,  at commit 49f21f6acad5ba95d71eb20d5eb0733c17f5eb14.

When tsd.storage.salt.width > 0,  set tsd.core.meta.enable_realtime_ts to true, the query '/api/search/lookup?m=sys.cpu&limit=10000000&type=*' will return 404 Not Found,  when tsd.storage.salt.width is 0, the query returns expected results successfully.   This bug makes Grafana can't show suggestions for tag keys and values.

```
# tsdb uid --config /etc/opentsdb/opentsdb-monitoring.conf sys.cpu
metrics sys.cpu: [100, 25, -15]

# curl 'http://localhost:4243/api/search/lookup?m=sys.cpu&limit=10000000&type=*' -s | /root/jsonpp.pl
{
   "error" : {
      "code" : 404,
      "details" : "No such unique ID for 'metrics': [25, -15, 0]",
      "message" : "Unable to resolve one or more UIDs",
[...]

PR #569, opened 2015-09-16
fix /api/search/lookup when salting is enabled

Dieken:
When salting is enabled, the row key is expected to be "salt + tsuid",
we can construct row key and still call RowKey.metricNameAsync(), but
it's a little more efficient to directly extact metric uid from tsuid
and call TSDB.getUidName().

This fixes issue https://github.com/OpenTSDB/opentsdb/issues/568.

PR #578, opened 2015-10-05
fix /api/search/lookup when salting is enabled

Dieken:
// this is to rebase #569 on latest branch 'next',  I can't update #569 because it's closed.

When salting is enabled, the row key is expected to be "salt + tsuid",
we can construct row key and still call RowKey.metricNameAsync(), but
it's a little more efficient to directly extact metric uid from tsuid
and call TSDB.getUidName().

This fixes issue https://github.com/OpenTSDB/opentsdb/issues/568.

### DIFF
--- a/src/tsd/SearchRpc.java
+++ b/src/tsd/SearchRpc.java
@@ -13,6 +13,7 @@
 package net.opentsdb.tsd;
 
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


manolama:
Merged in c4225ffbe32800b7e49e61d1ff8249a926bc983e

PR #581, opened 2015-10-07
Bug fixes for failing unit tests when UID width is changed from the default 3…

haridsv:
… (specifically tested with 4)

The following tests failed as of 2.1.0:
  TestTimeSeriesLookup.tagkOnlyMeta:211 expected:<5> but was:<6>
  TestTimeSeriesLookup.tagkOnlyMetaStar:227 expected:<5> but was:<6>
  TestTimeSeriesLookup.tagkOnlyData:244 expected:<5> but was:<6>
  TestTimeSeriesLookup.tagkOnly2Meta:260 expected:<2> but was:<0>
  TestUniqueIdRpc.tsuidPostByM


haridsv:
After changing the UID width, I had to hack the unit tests for the extra byte and then found that all except 5 unit tests still failed, so after debugging them, I found a couple of hardcoded assumption about the width being 3. There are also additional regex fixes in wildcards that I shouldn't be there and could have caused some corner case failures even with the 3-width UID's.

haridsv:
@manolama I signed the CLA, anything else I should do?

manolama:
Looks good, thanks a ton! Merged in e1d9ead07d689086f745f021436750458e98f598

haridsv:
Thanks a lot!

PR #582, opened 2015-10-07
enable /api/uid/rename endpoint

huttli:
As the title says, this pull request exposes the UID rename as an api endpoint.

This comes handy when we need to rename the tag value. Currently only one rename at one api request.

Sorry about the space removal. I wish github's pull request viewer has an option to turn them off. :D


huttli:
Closing this one for another clearer that doesn't have the unnecessary white spaces.

PR #586, opened 2015-10-09
Fix up Makefile.am

louxiu:
MAC OSX 10.10.5 (14F27)
automake (GNU automake) 1.15

I want to install opentsdb to a custom directory, say `install`.
`./build.sh ; ./configure --prefix=`pwd`/install; make; make install`  failed with output below.

/bin/sh: ./build-aux/install-sh: No such file or directory
make[2]: *** [install-data-local] Error 1
make[1]: *** [install-am] Error 2

Because `cd "$(DEV_TSD_STATICROOT)"` has change current directory. The commit can fix it.


manolama:
Thanks! Merged in 8dcd77d8907e3a0eaebde0eede81b2f76150dfd7

PR #587, opened 2015-10-09
Support environment variables to specify a custom DATA_BLOCK_ENCODING…

haridsv:
… and TTL


manolama:
Cool, could you validate which versions of HBase this works on? We'd still like to support at least 0.94+ and maybe even 0.92.

haridsv:
I haven't tested it on 0.94, but as far as I can tell both DATA_BLOCK_ENCODING and TTL are supported in latest 0.94 and the value FOREVER is valid for TTL, so this should work. I am not sure about 0.92.

PR #589, opened 2015-10-14
Increase max number of rows to be returned per Scanner round trip

johann8384:
We found that increasing this value improved our query time for many of our larger metrics (those with many tag values for a tag, or many tags).


manolama:
Awesome, thanks! Merged in ba636dc5de0adb2da9775d1afacf601f6b4002a9.

I changed the config name a bit to match the other HBase parameters and tweaked the method name.

PR #592, opened 2015-10-15
Allow to define special characters for keys and values of the tag via…

newpcraft:
… config tsd.core.tag.allow_specialchars

### DIFF
--- a/src/core/TSDB.java
+++ b/src/core/TSDB.java
@@ -208,6 +208,11 @@ public final class TSDB {
       uid_cache_map.put(TAG_VALUE_QUAL.getBytes(CHARSET), tag_values);
       UniqueId.preloadUidCache(this, uid_cache_map);
     }
+
+    if (config.getString("tsd.core.tag.allow_specialchars") != null) {
+      Tags.setAllowSpecialChars(config.getString("tsd.core.tag.allow_specialchars"));
+    }
+
     LOG.debug(config.dumpConfiguration());
   }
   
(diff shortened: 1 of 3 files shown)


manolama:
Thanks for this! I'd actually like to allow all printable and unicode characters (I have a patch for that I could send you). Could you also tweak this so it applies only to the tag values? I'd like to avoid special characters in the metrics and tag keys. Thanks!

newpcraft:
Yes, I agree as you mentioned that it must be avoid special characters in metrics and tag keys. I should have to handle them for the key and the value separately.
I will more improve this if you share your code for me, and I also try to consider whether we need my patch or not due to fully support all printable and unicode characters in value field.

manolama:
Would you mind signing https://docs.google.com/forms/d/1VVqKFTYNp3btgwXE51Zf9_Sfq66yZ4JbuEFrLlTD70E/viewform?embedded=true&formkey=dFNiOFROLXJBbFBmMkQtb1hNMWhUUnc6MQ so I can pull this in please? I'll probably change the name to something like "character_whitelist" for the config param. Thanks!

newpcraft:
Of course, I just signed it. Thanks.

johann8384:
@newpcraft If you can rebase from the latest "next", I'll give this a test-run and merge it.

newpcraft:
I created again a new one on the following PR. https://github.com/OpenTSDB/opentsdb/pull/780

PR #599, opened 2015-10-26
alignTimestamp() according to the TimeZone if setted.

inshion:
alignTimestamp() according to the TimeZone if setted. Or you cannot get the right result if you want to see datapoint aligned to a non-UTC timezone.

for example:
the query like:

    $ tsdb query 1445529600 1445788800 zimsum downsample 86400000 zimsum my.mtrc 
you'll get:

    ……
    my.mtrc 1445558400000 2570882.000000 {}
    my.mtrc 1445644800000 2591490.000000 {}
    my.mtrc 1445731200000 2879812.000000 {}
    ……
**1445558400000** is  **2015-10-23 00:00:00 UTC** 

but if your timezone is GMT+8， the actual result you want is:

    ……
    my.mtrc 1445529600000 2515360.000000 {}
    my.mtrc 1445616000000 2663943.000000 {}
    my.mtrc 1445702400000 2888318.000000 {}
    ……
**1445529600000** is  **2015-10-23 00:00:00 GMT+8**

### DIFF
--- a/src/core/Downsampler.java
+++ b/src/core/Downsampler.java
@@ -13,6 +13,7 @@
 package net.opentsdb.core;
 
 import java.util.NoSuchElementException;
+import java.util.TimeZone;
 
 
 /**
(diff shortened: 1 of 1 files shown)


manolama:
Thanks for taking a crack at this! Unfortunately this doesn't work because the numeric timestamps are in Unix Epoch format, which is UTC by definition.

There are a couple of options:
1) You can help us validate that at query time with an absolute timestamp (e.g. "2015-10-23 00:00:00") and time zone ("GMT+8" or "America/New York") that the TSD adjusts the time properly.
2) If you're querying with epochs, then adjust for the time zone before sending the epoch timestamp in the query.

Also, you could write a serializer that emits the timestamps as absolute values with time zones. Some users are interested in that and I'd be happy to include it.

Let me know if you have any questions and thanks!

PR #601, opened 2015-10-27
make OOM handler be able to handle multiple instances of opentsdb server

Dieken:
The "/etc/init.d/opentsdb" can be made multiple symlinks to launch multiple opentsdb instances, such as:
```bash
$ ls -l /etc/init.d/opentsdb* /etc/opentsdb/opentsdb* /etc/sysconfig/opentsdb*
lrwxrwxrwx 1 root root   39 Oct 27 13:46 /etc/init.d/opentsdb -> /usr/share/opentsdb/etc/init.d/opentsdb
lrwxrwxrwx 1 root root    8 Sep  8 16:25 /etc/init.d/opentsdb-monitoring -> opentsdb
lrwxrwxrwx 1 root root    8 Sep  8 16:25 /etc/init.d/opentsdb-search -> opentsdb
-rw-r--r-- 1 root root 2101 Oct 27 13:41 /etc/opentsdb/opentsdb.conf
-rw-r--r-- 1 root root 2878 Oct 27 04:01 /etc/opentsdb/opentsdb-monitoring.conf
-rw-r--r-- 1 root root 2862 Oct 27 04:01 /etc/opentsdb/opentsdb-search.conf
-rw-r--r-- 1 root root  187 Oct 27 13:43 /etc/sysconfig/opentsdb-monitoring
-rw-r--r-- 1 root root  183 Oct 27 13:44 /etc/sysconfig/opentsdb-search
```

But the script specified by "-XX:OnOutOfMemoryError=/usr/share/opentsdb/tools/opentsdb_restart.py" can only handle service "opentsdb",  this little patch enhances it to support multiple instances of opentsdb server.

### DIFF
--- a/tools/opentsdb_restart.py
+++ b/tools/opentsdb_restart.py
@@ -8,8 +8,11 @@ This is known to work with python2.6 and above.
 import os
 import subprocess
 
+service_name = "opentsdb"
+if 'NAME' in os.environ:
+    service_name = os.environ['NAME']
 
-subprocess.call(["service", "opentsdb", "stop"])
+subprocess.call(["service", service_name, "stop"])
 # Close any file handles we inherited from our parent JVM. We need
 # to do this before restarting so that the socket isn't held open.
 openfiles = [int(f) for f in os.listdir("/proc/self/fd")]
@@ -17,4 +20,4 @@ openfiles = [int(f) for f in os.listdir("/proc/self/fd")]
 # that there is less chance of errors with those standard streams.
 # Other files start at fd 3.
 os.closerange(3, max(openfiles))
-subprocess.call(["service", "opentsdb", "start"])
+subprocess.call(["service", service_name, "start"])


Dieken:
@manolama, could you merge this little patch, this looks impossible to introduce bug into 2.2 release:-)

I'm setting up a monitoring service with opentsdb as storage, really hate to keep my own fork for little feature:-D

I meant to use command argument but failed on space character in bash variable expansion in service start script, so I fell back to environment variable to keep change small.

manolama:
Thanks @Dieken! Since it's kind of a newish feature I'm putting it in 2.3 (the put branch) as 2.2 is in RC and locked down except for bug fixes. Committed in ce4660d643cc269da25d79f609b45b56b52bd896.

Dieken:
@manolama  May I beg your pardon to cherry-pick this into origin/next?  The /etc/init.d/opentsdb script supports multiple instances but opentsdb_restart.py doesn't,  so it's kind of a bug too :-)

Hope to not keep my private branch for v2.2, currently this is the only extra patch I need upon v2.2.

Dieken:
btw, @manolama,  is it ok to have multiple opentsdb instances to write same hbase tables for load balance?  Some days ago I found "tsdb fsck" reported lots of invalid rows, not sure whether it was due to #604 or concurrent compaction from multiple instances.  I haven't got a chance to recur, hope to double check in next weekend.

manolama:
@Dieken Since it's not a major tweak I went ahead and merged in 2.1.2 so you're good to go there.

Yep, it's perfectly fine to run multiple TSDs against HBase. We run about 150 TSDs in one of our colos against 110 region servers without an issue.

Dieken:
@manolama, thanks for your kindly merge:-)

That's a really big OpenTSDB installation, how do you set tsd.storage.enable_compaction and tsd.core.meta.enable_realtime_ts? A dump of call to /api/config would be very appreciated.

I ask this because in my last deployment the tsmeta table couldn't be scanned even with "hbase shell", it just hang. And I got corrupted row keys mentioned above. My config was paste at https://github.com/OpenTSDB/opentsdb/issues/568 except the hbase setting. I have 19 emitter machines each emits about 1000 rows every 10s, and 9 tsd instances, 1 region server(I have 10 but forgot to pre-split tsdata table, so actually used 1 region server for tsdata).

I will recur this weekend or early next week, currently the data center is frozen, can't deploy.

Dieken:
I plan to disable enable_compaction to avoid concurrent compaction, may use a cron job to call "tsdb fsck --compact FROM_TIME END_TIME".  And disable enable_realtime_ts to reduce load on tsmeta table, but it's a pity "tsdb uid metasync/treesync" doesn't support time range.

Any advice? Thanks!

manolama:
Hi @Dieken Here's our config, with some non-upstream (yet) tweaks https://gist.github.com/manolama/12d7f913575ed39d4c79

We disabled compactions and are just using puts right now. We may switch back to appends so that's something you may want to look into with 2.2. It writes data to a single column and avoids the need for compactions entirely. It does eat up more CPU on the HBase servers though so just monitor that if you want to try it out.

I also don't have real-time meta enabled though I do have a tweak that allows for buffering of that data and periodically flushing to HBase. I'll try to upstream that in 2.3 here.

And yeah, the tree sync can't easily support a time range because it scans the meta table which isn't time based any more.

Issue #571, opened 2015-09-18
Conf file specifies space separated list for ZK qurum


jregovic:
The conf file in the most recent 2.2 RPM specifies that the ZK quorum should be space separated, but spaces do not work. The comment should say comma separated.

manolama:
Fixed with the commit. Thanks!

PR #602, opened 2015-10-28
s/space/comma/ in instruction for specifying ZooKeeper quorum

nhahtdh:
Fix for issue https://github.com/OpenTSDB/opentsdb/issues/571

I grep the source folder, so we can be sure that all config files are updated.


manolama:
Merged in 13067fa9a20e2c8e35d147c17fade8a9a789eed4. Thanks!

PR #604, opened 2015-10-29
fix stuck metasync when salting is enabled

Dieken:
MetaScanner extracts (salt_width + metric_width) bytes from tsuid
to metric uid, this triggers IllegalArgumentException in
UniqueId.getNameAsync(id), then MetaScanner.call() won't never
call result.callback(null), and result.joinUninterruptibly() in
MetaSync.run() never returns.

BTW, I checked all usages of Const.SALT_WIDTH() and fixed some
other similar wrong calculations.


manolama:
Awesome, thanks! Merged in de0bceaa3b8ce4cb8137523a67eb391e42a5f790

PR #609, opened 2015-11-09
Fix IncomingDataPoint toString, separate tags with space

louxiu:
This is a minor fix to separate tags with space in `toString` of `IncomingDataPoint`.


manolama:
Merged in 3866e8c1b045ae900599bf09fffb7713ca881520. Thanks!

PR #621, opened 2015-11-17
Adding support for filtering metric names to be auto-created

johann8384:
Compares a comma delimited list of regular expressions to the presented metric name to decide if it should be created or not.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -75,7 +75,6 @@ tsdb_SRC := \
 	src/meta/UIDMeta.java	\
 	src/query/QueryUtil.java	\
 	src/query/expression/Absolute.java	\
-	src/query/expression/Expression.java	\
 	src/query/expression/DiffSeries.java	\
 	src/query/expression/DivideSeries.java	\
 	src/query/expression/EDPtoDPS.java	\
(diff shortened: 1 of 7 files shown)


johann8384:
I did some very basic testing in our environment.
## A good metric name:
2015-11-17 16:10:45,055 DEBUG [OpenTSDB I/O Worker #1] UniqueId: Accepted name for UID: testing.metrics.foo based on 'test..*'
2015-11-17 16:10:45,056 INFO  [OpenTSDB I/O Worker #1] UniqueId: Creating an ID for kind='metrics' name='testing.metrics.foo'

## A bad metric name:
2015-11-17 16:10:59,250 DEBUG [OpenTSDB I/O Worker #2] UniqueId: Rejected name for UID: tbadmetrics.foo
2015-11-17 16:10:59,250 INFO  [OpenTSDB I/O Worker #2] UniqueId: UID cannot be assigned, name is not acceptable: tbadmetrics.foo
2015-11-17 16:10:59,255 ERROR [OpenTSDB I/O Worker #2] RpcHandler: [id: 0xe1397817, /50.116.203.7:59102 => /50.116.203.7:4244] Unexpected exception caught while serving [null, tbadmetrics.foo, 1447798259, 0, domain=test]
java.lang.RuntimeException: UID cannot be assigned, name is not acceptable: tbadmetrics.foo

johann8384:
Made this from the wrong branch, new PR is #627

Issue #629, opened 2015-11-19
[2.2-rc3] tsdb uid is broken, can't create new metrics


rluta:
I'm running OpenTSDB 2.2 rc3 deb package on Ubuntu 14.04 configured with sharing.width=1 and random_uids.

I have inconsistent results from `tsdb uid`:

```
$ tsdb uid grep metrics test
metrics test.daily.bps: [-12, 48, 125]
metrics test.daily.bytes: [-30, -45, -2]
metrics test.daily.pkts: [16, -85, 108]
metrics test.daily.pps: [-77, -112, 96]

$ tsdb uid assign metrics test.new
Exception in thread "main" java.lang.IllegalArgumentException: No enum constant net.opentsdb.uid.UniqueId.UniqueIdType.metrics
	at java.lang.Enum.valueOf(Enum.java:236)
	at net.opentsdb.uid.UniqueId$UniqueIdType.valueOf(UniqueId.java:59)
	at net.opentsdb.tools.UidManager.assign(UidManager.java:364)
	at net.opentsdb.tools.UidManager.runCommand(UidManager.java:162)
	at net.opentsdb.tools.UidManager.main(UidManager.java:127)

$ tsdb uid assign METRIC test.new
METRIC test.new: [1, 4, 11]

$ tsdb uid grep METRIC test.new
METRIC test.new: [1, 4, 11]

$ tsdb uid grep metrics test
metrics test.daily.bps: [-12, 48, 125]
metrics test.daily.bytes: [-30, -45, -2]
metrics test.daily.pkts: [16, -85, 108]
metrics test.daily.pps: [-77, -112, 96]
```

In HBase, metric names created with METRIC actually appear in column 'name:METRIC', other metrics appear in column 'name:metrics'

manolama:
Arg yeah, I goofed that up. Patch coming. Notes on your PR too. Thanks for finding and fixing!

manolama:
Fixed in fb255884470597b3b4e8220b4250d75630099cdb

PR #630, opened 2015-11-19
Assign explicit string values to UniqueId enum

rluta:
This pull request fixes issue #629 by assigning explicit string values to UniqueId enum matching the TSDB class constants and implementing a equalsName method on the Enum to compare it with a string.

### DIFF
--- a/src/tools/UidManager.java
+++ b/src/tools/UidManager.java
@@ -361,7 +361,7 @@ final class UidManager {
                             final short idwidth,
                             final String[] args) {
     boolean randomize = false;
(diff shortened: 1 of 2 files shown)


manolama:
Thanks for the patch! This is the "right" way to do it but unfortunately some of the HBase qualifiers as you noted) are flipping between ''metric'' and ''metrics'' so this patch breaks that behavior. Instead there's an existing method ''UniqueId.stringToUniqueIdType()'' that I should have used before. Patch coming with that fix and props to you for finding it. Thanks!

PR #634, opened 2015-11-21
Silence stray logging output in TSMeta.call()

vitaliyf:
This line was added in ab276c823ee082bc216951e371062d1d019652d6 and based on indentation is likely an accidental commit.  It causes continuous spamming in our logs after upgrade.  Changing log level to "debug" to quiet it down.


manolama:
Merged in ad0ccad10bc49b5579ae71063250ef57be8f0b8e, thanks!

Issue #616, opened 2015-11-11
Can there be some configurable limit on connections TSD can make?


utkarshcmu:
This is a netstat output from v2.1.1. , just 5 minutes after starting the service.

tcp        0      0 :::4242                     :::*                        LISTEN      25756/java
tcp     1075      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.62.9:39944    ESTABLISHED 25756/java
tcp     1021      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.63.106:61647  ESTABLISHED 25756/java
tcp     1026      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.62.9:39844    CLOSE_WAIT  25756/java
tcp     1005      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.63.106:61446  CLOSE_WAIT  25756/java
tcp      799      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.63.106:61610  ESTABLISHED 25756/java
tcp      928      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.62.9:39887    ESTABLISHED 25756/java
tcp     1231      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.63.106:61420  CLOSE_WAIT  25756/java
tcp     1465      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.63.106:61603  ESTABLISHED 25756/java
tcp        1      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.63.106:61606  CLOSE_WAIT  25756/java
tcp     1362      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.63.106:61471  CLOSE_WAIT  25756/java
tcp      843      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.63.106:61560  ESTABLISHED 25756/java
tcp      830      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.62.9:39782    CLOSE_WAIT  25756/java
tcp     1376      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.63.106:61492  CLOSE_WAIT  25756/java
tcp     1151      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.63.106:61655  ESTABLISHED 25756/java
tcp     2440      0 ::ffff:10.241.63.134:4242   ::ffff:10.241.62.9:39804    CLOSE_WAIT  25756/java
[...]

PR #638, opened 2015-11-24
Limit  the total connections tsd can serve

juliennju:
As https://github.com/OpenTSDB/opentsdb/issues/616 metioned, when tsd is not able to keep pace with comming requests, total connections limit refusing new connections, help tsd not crash. And tsd may return ok with outside improvement.


utkarshcmu:
+1

johann8384:
Merging in #740

PR #639, opened 2015-11-24
"tsdb fsck --fix-all" wrongly clears data points when salting enabled

Dieken:
Const.SALT_WIDTH() is actually not constant value, it mustn't be called on class initialization.

### DIFF
--- a/src/tools/Fsck.java
+++ b/src/tools/Fsck.java
@@ -114,11 +114,11 @@ final class Fsck {
   final AtomicLong vle_fixed = new AtomicLong();
   
   /** Length of the metric + timestamp for key validation */
-  private static int key_prefix_length = Const.SALT_WIDTH() + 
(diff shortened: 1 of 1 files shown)


Dieken:
The original code will make key_prefix_length to be 7:  0 (SALT_WIDTH) + 3 (metrics_width) + 4 (timestamp_bytes), but actually SALT_WIDTH may be changed later due to config file loading.

manolama:
Awesome, good catch. Merged in 90265791efe4824dba012d485b6e8cea8e844c8e  Thanks!

Dieken:
@manolama  this patch is required for v2.2, not only v2.3,  I see you merged it into origin/put,  but not origin/next,  is this a mistake?

manolama:
Hmm, thought I merged it, woops! Will do.

PR #610, opened 2015-11-09
Fix staging and build on FreeBSD

xmj:
Make sure that the filesystem outside of DESTDIR isn't touched,
hence set java.util.prefs.userRoot to $(HOME) and make sure etc/
config files go to $(sysconfdir)/etc/opentsdb instead of
$(pkgdatadir)/etc/opentsdb.

PR #670, opened 2016-01-04
Changing sysconfigdir with prefix in Makefile.am

davidone:
configure includes a sysconfdir option that is not present in Makefile.am (a sysconfigdir option is present, instead). This leads to tsdb using always a /etc/opentsdb config path that in FreeBSD (and in other systems al well, probably) is not the right one.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -293,7 +293,7 @@ printdeps:
 # This is kind of a hack, but I couldn't find a better way to adjust the paths
 # in the script before it gets installed...
 install-exec-hook:
-	script=tsdb; pkgdatadir='$(pkgdatadir)'; configdir='$(sysconfigdir)/etc/opentsdb'; \
+	script=tsdb; pkgdatadir='$(pkgdatadir)'; configdir='$(prefix)/etc/opentsdb'; \
           abs_srcdir=''; abs_builddir=''; $(edit_tsdb_script)
 	cat tsdb.tmp >"$(DESTDIR)$(bindir)/tsdb"
 	rm -f tsdb.tmp


xmj:
This regression was introduced in my PR https://github.com/OpenTSDB/opentsdb/pull/610/files, 
my bad for not catching it earlier.

davidone:
Actually if you use pkgdatadir instead of sysconfdir:
```
/usr/ports/databases/opentsdb/work % grep -E 'pkgdata(.*)configdir' opentsdb-2.1.3/Makefile.in
        script=tsdb; pkgdatadir='$(pkgdatadir)'; configdir='$(pkgdatadir)/etc/opentsdb'; \
        script=tsdb; pkgdatadir='/usr/share/opentsdb'; configdir='/etc/opentsdb'; \
```
the confdir var in /usr/local/bin/tsdb will be populated as:
```
/usr/ports/databases/opentsdb/work % grep ^configdir stage/usr/local/bin/tsdb
configdir='/usr/local/share/opentsdb/etc/opentsdb'
```

that is not the correct path.

davidone:
Please do not merge, it need more test actually.

davidone:
I changed sysconfdir with prefix because sysconfdir is already including a trailing /etc, while prefix hasn't. Not completely sure if this will break on Ubuntu/RedHat systems.

Makefile.am
```diff
@@ -293,7 +293,7 @@ printdeps:
 # This is kind of a hack, but I couldn't find a better way to adjust the paths
 # in the script before it gets installed...
 install-exec-hook:
-	script=tsdb; pkgdatadir='$(pkgdatadir)'; configdir='$(sysconfigdir)/etc/opentsdb'; \
+	script=tsdb; pkgdatadir='$(pkgdatadir)'; configdir='$(prefix)/etc/opentsdb'; \
```
xmj: $(prefix)/etc is actually the intent of $sysconfdir

xmj: https://www.gnu.org/prep/standards/html_node/Directory-Variables.html

davidone: My understading is that:
```
$(sysconfig)/opentsdb
```

should work on GNU/Linux too but unfortunately I have no Linux boxes to test this.

xmj: This should work - $(sysconfdir)/opentsdb

manolama:
Checked it with the debian and rpm packages, looked good. Merged in 6d2102af7d7391b759698979cab9ae74bea15d80. Thanks!

PR #676, opened 2016-01-17
compare error of CompactionQueue.class

juliennju:
When metrics are salted, the start pos of timestamp should add Const.SALT_WIDTH().


manolama:
Fixed for Java 6 and merged in 663302459e4509a9d3aed12c67cf48add4111df9. Thanks for catching it!

PR #689, opened 2016-01-28
Allow parent-child branches have the same display name

teamurko:
Currently, a sequence of branches in parent-child relationship having the same display name are squashed into one branch, e.g. a.b.b.c.c.d is squashed into a.b.c.d. While reading code I haven't noticed any design reason for doing this. It would make more sense to preserve original metric name structure in a tree, in general. E.g. in my case some metrics can have consecutive repeated parts in metric name. Putting squashed branch B's children into B's parent children set is not convenient from browsing perspective.

### DIFF
--- a/src/tree/Branch.java
+++ b/src/tree/Branch.java
@@ -109,7 +109,7 @@ public final class Branch implements Comparable<Branch> {
 
   /** The path/name of the branch */
   private TreeMap<Integer, String> path;
-  
+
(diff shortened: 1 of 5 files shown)


johann8384:
There are all kinds of problems with merging this against the NEXT branch.

@teamurko please make this change against the next branch and we'll get it merged in.

PR #721, opened 2016-03-04
Added timeShift function

### LINKED ISSUE #175: support time-shift of a metric
Graphite supports a timeShift(metric, delta) function which adjusts the time shown by the delta.  This is particularly useful for week-over-week/etc comparisons, where you plot one metric against that same metric 7 days ago.

It shouldn't be hard to implement, you just subtract the delta from the timestamp you are looking for before fetching it from the database.

johann8384:
Fixes #175


manolama:
Merged in c7dee0184bbc89aa90b7688c6224ca6cf3a2430d. Need to add some UTs also :)

PR #735, opened 2016-03-17
Http latency stats plugin

eswdd:
As discussed on the mailing list earlier this week.

I've been looking at the http latency metrics recently and want to get more accurate percentiles (mainly for higher values) than what can be achieved with the current histogram approach.

Many companies use external aggregators like statsd for this kind of thing so have introduced an optional plugin api to emit the timing events to a plugin (which would process asynchronously) so not write to the same metrics.

It still requires some documentation to be written but allows users to configure both a global implementation and/or specific implementations for particular usages. If none is specified it drops back to using the current Histogram which implements the same interface now. I had to rejig various references to static member variables and generally replaced this with constructor injection which I feel is cleaner.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -363,6 +363,7 @@ test_plugin_SRC := \
   test/plugin/DummyPluginA.java \
   test/plugin/DummyPluginB.java \
   test/search/DummySearchPlugin.java \
+  test/stats/DummyLatencyStatsPlugin.java \
   test/tsd/DummyHttpSerializer.java \
   test/tsd/DummyHttpRpcPlugin.java \
   test/tsd/DummyRpcPlugin.java \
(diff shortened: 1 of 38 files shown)


johann8384:
Assigning to @manolama as he wanted to review this.

eswdd:
I think the flexibility in different plugins is overkill so I have an alternate impl without it which i've submitted in PR #769

eswdd:
I have an initial impl in https://github.com/eswdd/opentsdb-statsd-latency and just working on a doco PR now.

eswdd:
doco here: https://github.com/eswdd/opentsdb.net/tree/latency_stats_plugin

eswdd:
Closing this defunct PR as it was replaced by #769 which is itself defunct.

PR #769, opened 2016-04-11
Http latency stats plugin - simpler

eswdd:
Alternate impl of #735 with less flexibility in configuration

### DIFF
--- a/src/core/SaltScanner.java
+++ b/src/core/SaltScanner.java
@@ -25,6 +25,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import net.opentsdb.meta.Annotation;
 import net.opentsdb.query.filter.TagVFilter;
+import net.opentsdb.stats.LatencyStatsPlugin;
 import net.opentsdb.stats.QueryStats;
 import net.opentsdb.stats.QueryStats.QueryStat;
 import net.opentsdb.uid.UniqueId;
@@ -111,23 +112,28 @@ public class SaltScanner {
    * goes pear shaped. Make sure to synchronize on this object when checking
    * for null or assigning from a scanner's callback. */
   private volatile Exception exception;
-  
+  /** We do scans so we need to contribute to scan latency stats */
+  private LatencyStatsPlugin scanlatency;
+
   /**
    * Default ctor that performs some validation. Call {@link scan} after 
    * construction to actually start fetching data.
+   *
    * @param tsdb The TSDB to which we belong
    * @param metric The metric we're expecting to fetch
    * @param scanners A list of HBase scanners, one for each bucket
    * @param spans The span map to store results in
    * @param filters A list of filters for processing
+   * @param scanlatency For recording perceived hbase latency
    * @throws IllegalArgumentException if any required data was missing or
    * we had invalid parameters.
    */
   public SaltScanner(final TSDB tsdb, final byte[] metric, 
                                       final List<Scanner> scanners, 
                                       final TreeMap<byte[], Span> spans,
-                                      final List<TagVFilter> filters) {
-    this(tsdb, metric, scanners, spans, filters, false, null, 0);
+                                      final List<TagVFilter> filters,
+                                      final LatencyStatsPlugin scanlatency) {
+    this(tsdb, metric, scanners, spans, filters, false, null, 0, scanlatency);
   }
   
   /**
@@ -150,7 +156,8 @@ public class SaltScanner {
                                       final List<TagVFilter> filters,
                                       final boolean delete,
                                       final QueryStats query_stats,
-                                      final int query_index) {
+                                      final int query_index,
+                                      LatencyStatsPlugin scanlatency) {
     if (Const.SALT_WIDTH() < 1) {
       throw new IllegalArgumentException(
           "Salting is disabled. Use the regular scanner");
@@ -189,6 +196,7 @@ public class SaltScanner {
     this.delete = delete;
     this.query_stats = query_stats;
     this.query_index = query_index;
+    this.scanlatency = scanlatency;
   }
 
   /**
@@ -214,7 +222,7 @@ public class SaltScanner {
    */
   private void mergeAndReturnResults() {
     final long hbase_time = System.currentTimeMillis();
-    TsdbQuery.scanlatency.add((int)(hbase_time - start_time));
+    scanlatency.add((int)(hbase_time - start_time));
     long rows = 0;
 
     if (exception != null) {
--- a/src/stats/Histogram.java
+++ b/src/stats/Histogram.java
@@ -12,6 +12,9 @@
 // see <http://www.gnu.org/licenses/>.
 package net.opentsdb.stats;
 
+import com.stumbleupon.async.Deferred;
+import net.opentsdb.utils.Config;
+
 import java.util.Arrays;
 
 /**
@@ -35,7 +38,7 @@ import java.util.Arrays;
  * <p>
  * This class is not synchronized.
  */
-public final class Histogram {
+public final class Histogram extends LatencyStatsPlugin {
 
   /** Interval between each bucket for the linear part of the histogram. */
   private final short interval;
@@ -62,6 +65,14 @@ public final class Histogram {
   /** Buckets where we actually store the values. */
   private final int[] buckets;
 
+
+  /**
+   * Constructor, initialises the histogram with the standard values used for time latency measurements.
+   */
+  public Histogram() {
+    this(16000, (short) 2, 100);
+  }
+
   /**
    * Constructor.
    * @param max The maximum value of the histogram.  Any value greater
@@ -77,7 +88,7 @@ public final class Histogram {
    *   0 &lt;= cutoff &lt;= max
    * </pre>
    */
-  public Histogram(final int max,
+  Histogram(final int max,
                    final short interval, final int cutoff) {
     if (interval > max) {
       throw new IllegalArgumentException("interval > max! interval="
@@ -103,7 +114,35 @@ public final class Histogram {
       + 1];
   }
 
-  /**
+    @Override
+    public void initialize(Config config) {
+        // do nothing
+    }
+
+    @Override
+    public Deferred<Object> shutdown() {
+        return Deferred.fromResult(null);
+    }
+
+    @Override
+    public void start() {
+      
+    }
+
+    @Override
+    public String version() {
+        return "2.0.0";
+    }
+
+    @Override
+    public void collectStats(StatsCollector collector, String metricName, String xtratag) {
+        collector.record(metricName + "_50pct", percentile(50), xtratag);
+        collector.record(metricName + "_75pct", percentile(75), xtratag);
+        collector.record(metricName + "_90pct", percentile(90), xtratag);
+        collector.record(metricName + "_95pct", percentile(95), xtratag);
+    }
+
+    /**
    * Computes the logarithm base 2 (rounded up) of an integer.
    * <p>
    * This is essentially equivalent to
@@ -146,7 +185,7 @@ public final class Histogram {
   }
 
   /** Returns the number of buckets in this histogram. */
-  public int buckets() {
+  int buckets() {
     return buckets.length;
   }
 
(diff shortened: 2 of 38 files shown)


eswdd:
Doco here: https://github.com/eswdd/opentsdb.net/tree/latency_stats_plugin3

src/core/SaltScanner.java
```diff
@@ -111,23 +112,28 @@
...
    * @throws IllegalArgumentException if any required data was missing or
    * we had invalid parameters.
    */
   public SaltScanner(final TSDB tsdb, final byte[] metric, 
                                       final List<Scanner> scanners, 
                                       final TreeMap<byte[], Span> spans,
-                                      final List<TagVFilter> filters) {
-    this(tsdb, metric, scanners, spans, filters, false, null, 0);
+                                      final List<TagVFilter> filters,
+                                      final LatencyStatsPlugin scanlatency) {
```
manolama: Needs a 2.3 override ctor for backwards API compatibility.

src/stats/Histogram.java
```diff
@@ -77,7 +90,7 @@
    *   0 &lt;= cutoff &lt;= max
    * </pre>
    */
-  public Histogram(final int max,
+  Histogram(final int max,
```
manolama: Breaks API, leave it public.

src/stats/Histogram.java
```diff
@@ -146,7 +189,7 @@ static final int log2rounddown(int n) {
   }
 
   /** Returns the number of buckets in this histogram. */
-  public int buckets() {
+  int buckets() {
```
manolama: Same here, these all need to remain public.

src/tsd/RpcHandler.java
```diff
@@ -65,21 +67,14 @@
...
-  public RpcHandler(final TSDB tsdb) {
-    this(tsdb, RpcManager.instance(tsdb));
-  }
-  
+
+  /** Keep track of the latency of HTTP requests. */
+  private final LatencyStatsPlugin httplatency;
+  /** For when we're asked for graphs. */
+  private final GraphHandler graph_handler;
+
```
manolama: Where did the ctor go? Please restore for the public API.

src/stats/LatencyStats.java
```diff
@@ -0,0 +1,103 @@
...
+  /** Created instances */
+  private static ConcurrentHashMap<String, LatencyStatsPlugin> instances = new ConcurrentHashMap<String, LatencyStatsPlugin>();
+
+  /**
+   * Get the latency stats plugin for a given measurement point.
+   * @param config The configuration for this TSDB instance
+   * @param instanceName The named measurement point
+   * @param metricName The name of the metric to emit aggregations to
+   */
+  public static LatencyStatsPlugin getInstance(Config config, String instanceName, String metricName) {
```
manolama: Probably better to just pass in the TSDB object so we can pull out the config and possibly make additional calls in the future.
Also need Finals and TSDB style non-camel case param names.

src/stats/LatencyStats.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Get the latency stats plugin for a given measurement point.
+   * @param config The configuration for this TSDB instance
+   * @param instanceName The named measurement point
+   * @param metricName The name of the metric to emit aggregations to
+   * @param xtratag    Extra tags to use when emitting aggregations
+   */
+  public static LatencyStatsPlugin getInstance(Config config, String instanceName, String metricName, String xtratag) {
```
manolama: Ditto the finals, TSDB object and names.

src/stats/LatencyStats.java
```diff
@@ -0,0 +1,103 @@
...
+    // simple existence check
+    if (instances.containsKey(instanceName)) {
+      return instances.get(instanceName);
```
manolama: Avoid the double fetch:
```
LatencyStatsPlugin stats = instances.get(name);
if (stats != null) {
  return stats;
}
```

src/stats/LatencyStats.java
```diff
@@ -0,0 +1,103 @@
...
+              ret.getClass().getCanonicalName() + "] version: "
+              + ret.version());
+    } else {
+      ret = new Histogram(16000, (short) 2, 100);
+    }
+    
+    if (instances.putIfAbsent(instanceName, ret) == null) {
+      ret.start();
+    };
+    return instances.get(instanceName);
```
manolama: Clean this up a bit to make it faster:
```
LatencyStatsPlugin existing = instances.putIfAbsent(instance_name, ret);
if (existing != null) {
  ret.shutdown();
  return existing;
}
return ret;
```
We also probably don't want the ``start();`` method. plugins should perform their startup procedures in the ``initialize()`` method.

src/stats/LatencyStats.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  public static List<Deferred<Object>> shutdownAll() {
```
manolama: jdocs

src/stats/LatencyStats.java
```diff
@@ -0,0 +1,103 @@
...
+    for (LatencyStatsPlugin p : instances.values()) {
+      multi.add(p.shutdown());
+    }
+    return multi;
+  }
+
+
+  static void clear() {
+    instances.clear();
+  }
```
manolama: Jdocs and make sure to shutdown the objects. Don't leave them floating on the heap.

src/stats/LatencyStats.java
```diff
@@ -0,0 +1,103 @@
...
+    };
+    return instances.get(instanceName);
+  }
+
+  public static List<Deferred<Object>> shutdownAll() {
+    List<Deferred<Object>> multi = new ArrayList<Deferred<Object>>(instances.size());
+    for (LatencyStatsPlugin p : instances.values()) {
+      multi.add(p.shutdown());
+    }
+    return multi;
```
manolama: Also want to clear the implementation. We should never leave the objects in the map if we've shut them down. So you can probably dedupe this with ``clear()`` below.

src/stats/LatencyStatsPlugin.java
```diff
@@ -0,0 +1,81 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2015  The OpenTSDB Authors.
```
manolama: 2016

src/stats/LatencyStatsPlugin.java
```diff
@@ -0,0 +1,81 @@
...
+package net.opentsdb.stats;
+
+import com.stumbleupon.async.Deferred;
+import net.opentsdb.utils.Config;
+
+/**
+ * Used to calculate latency stats for a given operation. May work synchronously by collecting measurements and
+ * recording directly into a {@link StatsCollector} or may emit events to an external system which writes back
+ * via TSDB write channels (e.g. Telnet or HTTP).
+ */
```
manolama: @since 2.4

src/stats/LatencyStatsPlugin.java
```diff
@@ -0,0 +1,81 @@
...
+   * up properly. The TSD will then shutdown so the operator can fix the
+   * problem. Please use IllegalArgumentException for configuration issues.
+   *
+   * @param config The TSDB configuration
+   * @param metricName The name of the metric to emit aggregations to
+   * @param xtratag    Extra tags to use when emitting aggregations
+   * @throws IllegalArgumentException if required configuration parameters are
+   *                                  missing
+   */
+  public abstract void initialize(final Config config, String metricName, String xtratag);
```
manolama: Pass in the whole TSDB object, finals and TSD naming convention.

src/stats/LatencyStatsPlugin.java
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Called when this plugin is live. Calls to {@link #add(int)} will not be made
+   * before this method is called. Under race conditions it's possible this method will never
+   * be called on a given instance of this class.
+   */
+  public abstract void start();
```
manolama: Ug, as much as I don't like having to call it makes sense.

src/stats/LatencyStats.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Manages latency stats plugins for given measurement points. Supports differing plugin by measurement point as well as
+ * specifying a global default via config. If no config exists then this will return an {@link Histogram}. Ensures each 
+ * returned instance is a singleton.
+ */
+public class LatencyStats {
```
manolama: Actually, just roll this class into the LatencyStatsPlugin since you're just adding static methods.

src/stats/LatencyStatsPlugin.java
```diff
@@ -0,0 +1,81 @@
...
+   */
+  public abstract void collectStats(final StatsCollector collector);
+
+  /**
+   * Adds a value to be measured.
+   *
+   * @param value The value to add.
+   * @throws IllegalArgumentException may be thrown if the value given is negative.
+   */
+  public abstract void add(final int value);
```
manolama: Could you change this to ``long`` and add a ``double`` override? We can expand the histo implementations later.

src/stats/StatsCollector.java
```diff
@@ -101,24 +101,6 @@ public final void record(final String name,
...
   /**
-   * Records a number of data points from a {@link Histogram}.
-   * @param name The name of the metric.
-   * @param histo The histogram to collect data points from.
-   * @param xtratag An extra tag ({@code name=value}) to add to those
-   * data points (ignored if {@code null}).
-   * @throws IllegalArgumentException if {@code xtratag != null} and it
-   * doesn't follow the {@code name=value} format.
-   */
-  public final void record(final String name,
```
manolama: Don't remove a public API.

src/tsd/GraphHandler.java
```diff
@@ -69,20 +70,21 @@
...
+  private final LatencyStatsPlugin gnuplotlatency;
 
   /** Executor to run Gnuplot in separate bounded thread pool. */
   private final ThreadPoolExecutor gnuplot;
 
   /**
    * Constructor.
    */
-  public GraphHandler() {
+  public GraphHandler(LatencyStatsPlugin graphlatency, LatencyStatsPlugin gnuplotlatency) {
```
manolama: Ctor override for compatibility.

src/stats/LatencyStatsPlugin.java
eswdd: Does this mean it defo won't make 2.3?

src/tsd/HttpQuery.java
```diff
@@ -92,24 +90,22 @@
...
+  private final GraphHandler graphhandler;
+
   /**
    * Constructor.
    * @param request The request in this HTTP query.
    * @param chan The channel on which the request was received.
+   * @param graphhandler The graph handler to process png requests
    */
-  public HttpQuery(final TSDB tsdb, final HttpRequest request, final Channel chan) {
+  public HttpQuery(final TSDB tsdb, final HttpRequest request, final Channel chan, final LatencyStatsPlugin httplatency, GraphHandler graphhandler) {
```
manolama: Don't change the public API here. Instead just use the ``plugin.getIntsance(name);`` method.

src/tsd/GraphHandler.java
```diff
@@ -69,20 +70,21 @@
     = new AtomicInteger();
 
   /** Keep track of the latency of graphing requests. */
-  private static final Histogram graphlatency =
-    new Histogram(16000, (short) 2, 100);
+  private final LatencyStatsPlugin graphlatency;
```
manolama: Actually, leave all these static. Use your ``getInstance();`` calls to instantiate the plugins. Then you don't have to modify all of the collection code.

manolama:
Thanks for taking a stab at this. It looks good for the most part, just a few things around:
- Make sure to add overrides for public APIs  so we maintain backwards compatibility
- You can use the ``getInstance()`` method for instantiating histos instead of modifying all the classes to be non-static.
- Make sure to strip out the Unit Test formatting calls. Those make it really hard to read the diffs and see what changed.

eswdd:
@manolama I think we can safely close this, am looking at a new impl at work based on your comments which we'll target either 2.4 or 3.0 depending on what's sensible when it's ready.

manolama:
Cool, thanks! Getting closer with 3.0 to something stableish that works.

PR #771, opened 2016-04-13
Typo: tsd.network.keepalive is actually called tsd.network.keep_alive according to code and documentation.

vitaliyf:
To match https://github.com/OpenTSDB/opentsdb/blob/master/src/utils/Config.java#L477


johann8384:
I think I already changed this in "next", I'll verify.

PR #774, opened 2016-04-15
Fix #773 - Add global_annotations to graphs

### LINKED ISSUE #773: Add global annotations to graphs
GraphHandler is currently not retrieving (or setting) global annotations when these are requested. All graphs ignore the queryParam "&global_annotations" and the gnuplot result is always missing the vertical lines (as it isn't retrieved or set).

I've fixed this (PR to follow) in GraphHandler.java because the codebase is actually handling queries differently in GraphHandler.java and QueryRpc.java (the code in GraphHandler is actually synchronous right now). In my opinion, both GraphHandler.java and QueryRpc.java should be refactored to separate query logic from the actual visualization (and thus, share the query logic code), but that's a much larger piece of work (we should probably open an issue for this).

HugoMFernandes:
Fixed a bug in which global_annotations were ignored by
GraphHandler.java (and thus, were not rendered by gnuplot). Added async
call to hbase to retrieve global annotations when required in a similar
fashion as that of QueryRpc.java.

Note: Since both GraphHandler.java and QueryRpc.java implement (similar)
query-processing logic, all of this should be unified in the future
(requires a large refactor to split all query logic with visualization
logic).

### DIFF
--- a/src/tsd/GraphHandler.java
+++ b/src/tsd/GraphHandler.java
@@ -45,10 +45,14 @@ import net.opentsdb.core.Query;
 import net.opentsdb.core.TSDB;
 import net.opentsdb.core.TSQuery;
 import net.opentsdb.graph.Plot;
+import net.opentsdb.meta.Annotation;
 import net.opentsdb.stats.Histogram;
 import net.opentsdb.stats.StatsCollector;
 import net.opentsdb.utils.DateTime;
 import net.opentsdb.utils.JSON;
+
+import com.stumbleupon.async.Callback;
+import com.stumbleupon.async.Deferred;
 
 /**
  * Stateless handler of HTTP graph requests (the {@code /q} endpoint).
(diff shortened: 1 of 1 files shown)


johann8384:
Is this a duplicate of #736?

Also, this will need to be rebased off "next" not "master".

HugoMFernandes:
No. Also, I'm not sure on how you tag things, but this is a bug fix ("feature request" seems weird).

Regarding master/next: Should I close this one and reopen with `next`? I read the contributing section (before contributing) and didn't see anything about it (it's probably a good idea to add it).

johann8384:
I can change the tags, bug or feature, just different way of thinking about
it. Bug, in that it isn't doing something you expect it to do, feature in
that the fix adds a feature that did not exist before and does now.

johann8384:
@HugoMFernandes Have you submitted the Contributor License Agreement which is linked to from here? http://opentsdb.net/docs/build/html/development/index.html

HugoMFernandes:
@johann8384 Had forgotten about it. I've done it now.

Issue #760, opened 2016-04-04
check_tsd references unknown attribute; unusable


stromnet:
tools/check_tsd line 231 tries to read options.bad_percent, which was removed in 010ed96a572f33b35b570cc4c4ce80a5a97b371a:

```
$ python /usr/local/share/opentsdb/tools/check_tsd -H xxxxx -p 4242 -m owfs.reading -t alias=device_cnt -w 34 -c 0 -v
Connected to 2001:xxxxxx:2:4242
send: 'GET /q?start=600s-ago&m=sum:owfs.reading{alias=device_cnt}&ascii&nagios HTTP/1.1\r\nHost: back-1.stromnet.se:4242\r\nAccept-Encoding: identity\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Content-Type: text/plain
header: Age: 0
header: Cache-Control: max-age=0
header: Content-Length: 4256
owfs.reading 1459792803 34 alias=device_cnt bus=main type=stats_bus ch=None
owfs.reading 1459792863 34 alias=device_cnt bus=main type=stats_bus ch=None
...
owfs.reading 1459796050 34 alias=device_cnt bus=main type=stats_bus ch=None
owfs.reading 1459796110 34 alias=device_cnt bus=main type=stats_bus ch=None

ignored 46/56 data points for being more than 600s old
worse data point value=34 at ts=1459795569
Traceback (most recent call last):
  File "/usr/local/share/opentsdb/tools/check_tsd", line 258, in <module>
    sys.exit(main(sys.argv))
  File "/usr/local/share/opentsdb/tools/check_tsd", line 231, in main
    if options.bad_percent is not None and rv > 0 \
AttributeError: Values instance has no attribute 'bad_percent'
```

check_tsd can thus never output a valid OK value, it crashes before it reaches the OK printer.

Also described in #728.

stromnet:
Fix available in PR #776

IDerr:
@manolama you can close this one, merged here : https://github.com/OpenTSDB/opentsdb/pull/779

PR #776, opened 2016-04-16
Misc check_tsd fixes & improvements

stromnet:
First of all, this unbreaks #760, making the tool usable again.

It also adds a new feature, running with -N <utc timestamp> for testing against historic data, rather than just the latest data. This is very useful when tweaking parameters such as downsampling, aggregation etc, against historic data. In my case I've used it to find out which parameters would be good to trigger alarm on a temperature sensor which has a normal level, but regularly have some short peaks which I wanted to ignore.
While tweaking, I found out that one of the new downsample algorithms worked best. So I've added all the ones listed on http://opentsdb.net/docs/build/html/user_guide/query/aggregators.html#available-aggregators, both for use as downsampler and aggregator. Also made the verbose output explicitly show what it ignored.

### DIFF
--- a/tools/check_tsd
+++ b/tools/check_tsd
@@ -29,6 +29,13 @@ import sys
 import time
 from optparse import OptionParser
 
+AGGREGATORS = ('avg', 'count', 'dev',
+        'ep50r3', 'ep50r7', 'ep75r3', 'ep75r7', 'ep90r3', 'ep90r7', 'ep95r3', 'ep95r7',
+        'ep99r3', 'ep99r7', 'ep999r3', 'ep999r7',
+        'mimmin', 'mimmax', 'min', 'max', 'none',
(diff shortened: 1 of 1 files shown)


johann8384:
Two questions for @stromnet:
1) Have you submitted the CLA previously? http://opentsdb.net/docs/build/html/development/index.html
2) Have you submitted the corresponding documentation update? https://github.com/opentsdb/opentsdb.net

stromnet:
@johann8384: 
# CLA now signed!
# I have not updated the docs. That would be https://github.com/OpenTSDB/opentsdb.net/blob/gh-pages/docs/source/user_guide/utilities/nagios.rst, right?

johann8384:
yeah, that looks right. Those docs probably need and overall refresh, update it with your new functionality and I'll save the refresh for another day.

stromnet:
PR for doc above

johann8384:
Merging in #779

PR #785, opened 2016-04-25
Fix comment for behavior of tsd.network.tcp_no_delay

kev009:
And catch the rpm and deb confs up a bit.

I'm also not sure if the "Default is True" for these networking options is correct.  Does i.e. https://github.com/OpenTSDB/opentsdb/blob/e2efe9a4875f8cddd3404e9c6b64f5aab4b72c0a/src/tools/TSDMain.java#L171 default to True?


manolama:
Cool, yeah the default is true at https://github.com/OpenTSDB/opentsdb/blob/master/src/utils/Config.java#L476

manolama:
merged in 202efa07e4049927206c4e24317567cb0971d21b

PR #791, opened 2016-05-10
Temporary blacklisting of queries that ends up scanning large number of rows

r-santhosh:
We use opentsdb as our metric system for application and system health monitoring.

Some times, we end up having a scenario in which one of region server becomes hot spotted due to a query that scans large number of rows. This is due to the high cardinality for the metric and time range for which the query is issued.

To mitigate this scenario, we've implemented a temporary blacklisting on such queries. The code change is attached as a patch file with this email.

This change has been done in the query layer because it has the right contextual information like rows scanned to best take the decision about blacklisting. This code change will reactively blacklist queries that scan more than 'n' rows for a specific duration of time 't'. This feature can be turned off/on via config and it's off by default. Settings like max number of records ('n'), blacklist interval ('t') is configurable via config file.

Currently this change works only when the query is made using the metric name and does not handle the scenario where the query is based on TSUID.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -146,6 +146,7 @@ tsdb_SRC := \
 	src/tsd/AbstractHttpQuery.java	\
 	src/tsd/AnnotationRpc.java	\
 	src/tsd/BadRequestException.java	\
+	src/tsd/BlacklistRpc.java	\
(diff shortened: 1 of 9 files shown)


chinausers:
ok?

Issue #823, opened 2016-06-30
Potential infinite loop from multithreaded use of HashMap by way of "skip" HashSet in SaltScanner?


thatsafunnyname:
Hello and thank you for opentsdb,

We are using `opentsdb-2.2.0` with

```
tsd.storage.salt.buckets = 30
tsd.storage.salt.width = 1
```

and `mapr-hadoop-core-2.7.0.37549.GA-1.x86_64`, both using `java-1.7.0-openjdk-1.7.0.101.x86_64` on `RHEL6.6` on `2.6.32-573.12.1.el6.x86_64`

We see the opentsdb process occasionally spinning CPU in user time, logs from querying the `api/stats/threads` HTTP interface show a thread resizing a `HashMap` at these times.  We do not have a reproducer.  We have a monit process that detects the high CPU and restarts opentsdb.
 
I read http://mailinator.blogspot.co.uk/2009/06/beautiful-race-condition.html and was wondering if we are hitting this race, and if this might explain what we see.
 
I will attach the `api/stats/threads` log files.  Some process stats for opentsdb process (the host has 12 vCPUS):

``` 
# TOP PROCESSES sorted by time (counters are /sec) 16:57:00
# PID  User     PR  PPID THRD S   VSZ   RSS CP  SysT   UsrT Pct  AccuTime  RKB  WKB MajF MinF 
96659  opentsdb 20     1  213 S   15G  719M  7  0.01  0.12   0   0:24.21    0    1    0    8

# TOP PROCESSES sorted by time (counters are /sec) 16:58:00
# PID  User     PR  PPID THRD S   VSZ   RSS CP  SysT   UsrT Pct  AccuTime  RKB  WKB MajF MinF 
96659  opentsdb 20     1  213 S   16G    1G  7  1.01 179.33 300   3:24.55    0   30    0 4227

# TOP PROCESSES sorted by time (counters are /sec) 16:59:00
# PID  User     PR  PPID THRD S   VSZ   RSS CP  SysT   UsrT Pct  AccuTime  RKB  WKB MajF MinF 
96659  opentsdb 20     1  213 S   16G    1G  7  0.65 415.73 693  10:20.93    0    1    0    8

# TOP PROCESSES sorted by time (counters are /sec) 17:00:00
# PID  User     PR  PPID THRD S   VSZ   RSS CP  SysT   UsrT Pct  AccuTime  RKB  WKB MajF MinF 
[...]

PR #835, opened 2016-07-07
Use thread-safe ConcurrentHashMap. (#3)

thatsafunnyname:
For https://github.com/OpenTSDB/opentsdb/issues/823
Concurrent use of a HashSet can trigger race conditions and infinite loop(s), use the thread-safe ConcurrentHashMap to avoid this.
Making the change just for the thread stack seen looping.


manolama:
Merged in 6582775dbca7672d22297918af613d7004fd65ff. Thanks!

PR #838, opened 2016-07-13
Log X-Forwarded-For address when handling HTTP requests

CrackerJackMack:
We had an issue where we couldn't track source IP's through the load balancer as the netty Channel in the AbstractHttpQuery was only logging the connected socket IP.  This patch enhances the AbstractHttpQuery class to log the real client IP, if available in the headers, in a mostly compatible format to [AbstractChannel.toString()](http://netty.io/4.0/api/io/netty/channel/AbstractChannel.html#toString()).

Seems to work w/o issue so far, the only _issue_ I see right now is that I can't reliably put a port after the real client IP.  It's superfluous to do so as we aren't talking directly with the client, but we we need to be able to track down connections via channel ID as well as loadbalancer IP and remote IP.

This should be compatible with any frontend proxies (such as varnish, nginx, apache, haproxy, etc) as well as commercial platforms that follow the ["standard"](https://en.wikipedia.org/wiki/X-Forwarded-For).  This does not take into consideration [RFC7239](https://tools.ietf.org/html/rfc7239) which uses the new `Forwarded` header.

### DIFF
--- a/src/tsd/AbstractHttpQuery.java
+++ b/src/tsd/AbstractHttpQuery.java
@@ -467,21 +467,37 @@ public abstract class AbstractHttpQuery {
     return LOG;
   }
 
+  protected final String logChannel() {
+    if (request.containsHeader("X-Forwarded-For")) {
+        String inetAddress;
+        String proxyChain = request.getHeader("X-Forwarded-For");
+        int firstComma = proxyChain.indexOf(',');
+        if (firstComma != -1) {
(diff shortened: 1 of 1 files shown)


CrackerJackMack:
Forgot to mention I lifted a small bit of the X-Forwarded-For extraction from https://gist.github.com/jbowes/1065230.  Seemed trivial and did the right thing, if that's a problem lemme know and I can re-write the extraction method.

Issue #851, opened 2016-08-24
Add new Aggregator , which puts previous value at timestamp with missing data 


SumanjeetBhatti:
In OpenTSDB,  instead of doing linear Interpolation at missing data points, add previous value at the time-stamp with missing data. Then do the sum of all the non-aligned time series into single time-series.
For eg.
           t0   t0+5  t0+10 t0+15 t0+20 t0+25 t0+30
A         10              20                  15               25
B         12    20                  30                         25
Sum    22    30      40        50       45     45      50
So that for A ,at t0+5 value gets 10, at t0+15 value gets 20 and at t0+25 value gets 15.
For B ,at t0+10 value gets 20,at t0+20 value gets 30 and at t0+25 again 30.

SumanjeetBhatti:
Added a new aggregator "pfsum".

johann8384:
Fixed with #854 and #858

PR #854, opened 2016-08-31
#851- Added new aggregator to add previous value,in case data point is missing 

SumanjeetBhatti:
Added new aggregator to add previous value,in case data point is missing and then sum up all data points to single metrics.


johann8384:
Merging in #854

PR #858, opened 2016-09-01
Jcreasy/adjust 854

johann8384:


manolama:
Yeah, in this case we have the "fillPolicy" that can write NaNs, Nulls or Zeros. (need to allow for other scalars as well, as we do in Expressions). 

@SumanjeetBhatti What's your use-case for this fill? Thanks!

johann8384:
Not merging the Aggregator, should use a fill-policy

SumanjeetBhatti:
FillPolicy works with downsampling in query.
Our use case is we need each timestamp value,as it is saved in TSDB. No collaboration or downsampling .

That's why I added a new aggregator in the code.
If I can achieve my requirement with fill policy and without downsampling, let me know.

PR #859, opened 2016-09-05
add bad percent option

### LINKED ISSUE #728: check_tsd error
Hi,

I am getting following error while running check_tsd  for setting up a nagios alert  - 

matrix - http.hits 1458021000 450 host=A

**worse data point value=450 at ts=1458021000**
Traceback (most recent call last):
  File "./check_tsd", line 258, in <module>
    sys.exit(main(sys.argv))
  File "./check_tsd", line 231, in main
    if options.bad_percent is not None and rv > 0 \
AttributeError: Values instance has no attribute 'bad_percent'

Kindly tell me the solution for the same. Whatever  value it;s getting from the result metrics, it';s giving the same error "worst data point value".

mengzhuo:
fix #728


johann8384:
Merged in #861

PR #870, opened 2016-09-19
RPCManager, plugin loader refactor

### LINKED ISSUE #830: OpenTSDB Drops caches and returns a METHOD_NOT_ALLOWED error with DELETE

johann8384:
* Fixed inconsistency with dropCache endpoint
*  Added configuration options tsd.core.enable_api, tsd.core.enable_ui
* Refactored RPCManager, allowed for disabling the built-in UI or API via configuration options

Fix #830

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -98,7 +98,7 @@ tsdb_SRC := \
 	src/query/expression/PostAggregatedDataPoints.java	\
 	src/query/expression/Scale.java	\
 	src/query/expression/SumSeries.java	\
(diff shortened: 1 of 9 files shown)


johann8384:
The purpose of this is to reduce the number of RPC endpoints registered when they are not needed. There are cases where we wanted to drop all of the UI elements and the GraphHandler RPC so I added the --disable-ui flag for this purpose.

I also did some refactoring of the RPC Plugin loader in order to make it clearer what is where.

PR #871, opened 2016-09-20
Added Dockerfile and tool to build docker root

johann8384:
I use these to build read-only query nodes for use with [turn/splicer](https://github.com/turn/splicer).


bobrik:
FYI: https://github.com/cloudflare/docker-opentsdb

johann8384:
Yes, there are several variations of these.

PR #897, opened 2016-11-29
add aggregator Last()

EaconTang:
add aggregator Last()

Last(): get the last one data on the timestamp

### DIFF
--- a/src/core/Aggregators.java
+++ b/src/core/Aggregators.java
@@ -21,7 +21,6 @@ import org.apache.commons.math3.stat.descriptive.rank.Percentile.EstimationType;
 import org.apache.commons.math3.util.ResizableDoubleArray;
 
 import com.google.common.base.Preconditions;
-
 /**
  * Utility class that provides common, generally useful aggregators.
  */
@@ -34,7 +33,7 @@ public final class Aggregators {
     LERP,   /* Regular linear interpolation */
     ZIM,    /* Returns 0 when a data point is missing */
     MAX,    /* Returns the <type>.MaxValue when a data point is missing */
-    MIN     /* Returns the <type>.MinValue when a data point is missing */
+    MIN,    /* Returns the <type>.MinValue when a data point is missing */
   }
   
   /** Aggregator that sums up all the data points. */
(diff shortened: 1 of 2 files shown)


src/core/Aggregators.java
```diff
@@ -34,7 +33,7 @@
     LERP,   /* Regular linear interpolation */
     ZIM,    /* Returns 0 when a data point is missing */
     MAX,    /* Returns the <type>.MaxValue when a data point is missing */
-    MIN     /* Returns the <type>.MinValue when a data point is missing */
+    MIN,    /* Returns the <type>.MinValue when a data point is missing */
```
johann8384: Is this change intentional?

johann8384: Is this change intentional?

EaconTang: No, that is unconscious change...
Sorry for my carelessness, shall I just modified it and send another pull request to the "next" branch?

manolama: Hi @EaconTang yeah if you could revert that then squash all your commits into one please, that'd be great.

EaconTang: @johann8384 @manolama Hey， I just found that "first" and "last" aggregators already exist in "next" branch, and the implementation code makes me a little confused...

EaconTang: https://github.com/OpenTSDB/opentsdb/blob/next/src/core/Aggregators.java

Seems like the codes in class "First" and class "Last"  are almost the same?

manolama: Ah right yeah these are already in there at https://github.com/OpenTSDB/opentsdb/blob/next/src/core/Aggregators.java#L666-L708 in 2.3.0. I'll close this PR then.

The code for those aggregators is correct. The "First" aggregator fetches the first value from the iterator then iterates over the others to consume the iterator but doesn't assign the output anywhere. The "Last" aggregator iterates similarly but always assigns the iteration value to the variable.

manolama:
Duplicate of what's in https://github.com/OpenTSDB/opentsdb/blob/next/src/core/Aggregators.java#L666-L708

PR #930, opened 2017-02-10
add difference of between the first value and the last value for aggregators

c61811:
hello，please add this func for opentsdb，that used to calculate difference of between the first value and the last value in the data points for aggregators.
thanks!

### DIFF
--- a/src/uid/UniqueId.java
+++ b/src/uid/UniqueId.java
@@ -84,7 +84,7 @@ public final class UniqueId implements UniqueIdInterface {
   /** Maximum number of results to return in suggest(). */
   private static final short MAX_SUGGESTIONS = 25;
   /** Maximum number of  cache_hits. */
-  private static final int MAX_CACHE_SIZE = 1500000000;
+  private static final long MAX_CACHE_SIZE = 2000000000L;
 
   /** HBase client to use.  */
   private final HBaseClient client;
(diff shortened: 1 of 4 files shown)


c61811:
please give more advice someone.  this change is there a problem?

c61811:
I hope that the maintainer to seriously consider this feature, because in our scenario, the use of a metric values will continue to increase, but as we look at the index in every changes over a period of time. 

thanks!

johann8384:
This seems like it would be very similar to the "rate" functionality, what is this doing that I have overlooked?

c61811:
image
For example, the AppsSubmitted metric has been growing in Hadoop, we're going to see the specific changes of data of every time.

johann8384:
Isn't this what the "rate" functionality does? Converts a monotonically increasing counter to the derivative?

c61811:
Otherwise, The Int type of uid.cache-hit overflow, I hope to merge into the master branch.

image

c61811:
Isn't the Rate functionality, Although very similar, but the Diff functionality is more intuitive to us.

src/uid/UniqueId.java
```diff
@@ -249,6 +251,17 @@ public void dropCaches() {
   }
 
   /**
+   * Due to the var cache_hits type is int, but the max of int is 2147483648, and int happen spilling
+   */
+  private void reNumCache() {
```
manolama: I'd rather just change the hits to a long instead of rolling over this low.

c61811: hello @manolama, Please pay attention to the tsd.uid.ids-available is minus #1123, How to fix the ids-available that is positive? 
thanks!

manolama:
It does make sense to use the rate mostly, particularly since a bucket may wind up wit a single value. But some folks might like this. Merged in 26355ac18a2982f17edf773ab7ba711f085ac024

Issue #667, opened 2015-12-26
Opentsdb Security


MohamedHz:
I am successfully running the opentsdb with port 4242 and gathering the metrics from system. But  the issue we are facing practically was and who-ever can gathered the metrics from TSDB server by using port 4242 and draw the graphs.

Please share me the good security tips for securing the port 4242 and metrics should be gathered if only by authenticating the port by any system like ( For eg : Graphana ) not by simply typing the url : 192.168.1.1:4242

Awaiting for your reply!

trixpan:
@MohamedHz 

OpenTSDB has no security built in. Whoever can read it will also be able to write to it.

Cheers

mgoralczyk-viasat:
If you have 2.1+ you can set the tsd.mode setting to "ro" to make a TSD read-only, then with firewall settings make only the read-only TSDs accessible to the right people (or servers) for reading.

pau:
We delegate security to Nginx and use Basic HTTP Authentication.

```
location /tsdb/ {
        auth_basic "Restricted";
        auth_basic_user_file /etc/nginx/tsdb.htpasswd;
        proxy_pass http://localhost:4242/;
    }
```

Note that you should set `tsd.network.bind` to `127.0.0.1` so external requests to port 4242 can't get thru.

IDerr:
This one is resolved @manolama #899
You can close this :D

PR #958, opened 2017-03-27
Tags via custom HTTP header - provide some authentication

pladen:
Hi there,

Here is a PR we're using on a multi-tenant TSDB cluster (2.2.1). It allows to give tags to TSDB API via custom HTTP headers.

It basically allow to authenticate HTTP client on reverse proxy (nginx etc) with various mechanisms (SSL cert, IP address, basic auth, even SAML) and to push this info to TSDB within a custom header. This aims to securely separate data between our customers, given that the reverse proxy will properly deal with authentication.

You have to put the name of the header in your config, such as :
`tsd.http.header_tag = X-TSDB-TAG`

Then if you push data via the /api/put endpoint, just add key/value tags in this header :
`X-TSDB-TAG: customer=john,dc=virginia`

These tags and values will be added to the tags already present in JSON payload.
Example with haproxy and SSL client auth :
`http-request set-header X-TSDB-TAG customer=ssl_c_s_dn(cn)`

I am aware that some people are currently working on a real authentication framework (see #667). This patch does not aim to deal with user authentication but to permit customer isolation via third party authentication (ie: http reverse proxy).

We can write some docs about this, given that this PR is on the right track.

Note1 : tags in headers only work on "put" endpoint. It should be easy to use it also in "query" endpoint
Note2 : this PR is against next, but you can find a commit against master on our fork

Regards

### DIFF
--- a/src/tsd/PutDataPointRpc.java
+++ b/src/tsd/PutDataPointRpc.java
@@ -278,6 +278,7 @@ class PutDataPointRpc implements TelnetRpc, HttpRpc {
       throw new BadRequestException("No datapoints found in content");
     }
 
+    final HashMap<String, String> query_tags = new HashMap<String, String>();
     final boolean show_details = query.hasQueryStringParam("details");
     final boolean show_summary = query.hasQueryStringParam("summary");
     final boolean synchronous = query.hasQueryStringParam("sync");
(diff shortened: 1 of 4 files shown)


src/tsd/PutDataPointRpc.java
```diff
@@ -292,7 +293,18 @@ public void execute(final TSDB tsdb, final HttpQuery query)
     int queued = 0;
     final List<Deferred<Boolean>> deferreds = synchronous ? 
         new ArrayList<Deferred<Boolean>>(dps.size()) : null;
-    
+
+    if (tsdb.getConfig().enable_header_tag()) {
```
johann8384: For consistency, please use this method for checking the config value: 

 tsdb.getConfig().getString("tsd.http.enable_header_tag").equals("true");

src/utils/Config.java
```diff
@@ -564,6 +577,7 @@ protected void setDefaults() {
     default_map.put("tsd.core.stats_with_port", "false");    
     default_map.put("tsd.http.show_stack_trace", "true");
     default_map.put("tsd.http.query.allow_delete", "false");
+    default_map.put("tsd.http.header_tag", "");
```
johann8384: Please use: default_map.put("tsd.http.header_tag", "false");

pladen: The "header_tag" config parameter is actually a string, representing the name of the HTTP header to look for.
ie `tsd.http.header_tag  = X-TSDB-TAG`

Do you suggest to add a header_tag_enable boolean config parameter to enable/disable this behaviour ?

src/tsd/PutDataPointRpc.java
pladen: There is actually no "tsd.http.enable_header_tag" config parameter. See other portion of patch, i add a method to check the !null of "tsd.http.header_tag" value.

Maybe i should check this is the constructor, like how "send_telnet_errors" is managed ?

manolama:
I think it's ok, we'd need some UTs around it as well please.

johann8384:
I'll look at the issue with #1003 and merge this there once it's sorted.

PR #971, opened 2017-04-20
Tiered compaction

karanmehta93:
Made corresponding code changes
1. Added OpenTSDB Timestamp as HBase Cell Timestamp while writing a metric
2. For OpenTSDB compaction, it will use the maximum Timestamp out of all the stamps being compacted and use it to write for HBase Cell Timestamp.
3. Adding TS to scanner.

### DIFF
--- a/src/core/CompactionQueue.java
+++ b/src/core/CompactionQueue.java
@@ -86,6 +86,8 @@ final class CompactionQueue extends ConcurrentSkipListMap<byte[], Boolean> {
 
   /** If this is X then we'll flush X times faster than we really need.  */
   private final int flush_speed;  // multiplicative factor
+  
+  private long compactedKVTimestamp;
 
   /**
    * Constructor.
@@ -336,6 +338,7 @@ final class CompactionQueue extends ConcurrentSkipListMap<byte[], Boolean> {
         return null;
       }
 
+      compactedKVTimestamp = Long.MIN_VALUE;
       // go through all the columns, process annotations, and
       heap = new PriorityQueue<ColumnDatapointIterator>(nkvs);
       int tot_values = buildHeapProcessAnnotations();
@@ -385,7 +388,7 @@ final class CompactionQueue extends ConcurrentSkipListMap<byte[], Boolean> {
       deleted_cells.addAndGet(to_delete.size());  // We're going to delete this.
       if (write) {
         written_cells.incrementAndGet();
-        Deferred<Object> deferred = tsdb.put(key, compact.qualifier(), compact.value());
+        Deferred<Object> deferred = tsdb.put(key, compact.qualifier(), compact.value(), compactedKVTimestamp);
         if (!to_delete.isEmpty()) {
           deferred = deferred.addCallbacks(new DeleteCompactedCB(to_delete), handle_write_error);
         }
@@ -434,6 +437,7 @@ final class CompactionQueue extends ConcurrentSkipListMap<byte[], Boolean> {
           if (qual[0] == Annotation.PREFIX()) {
             annotations.add(JSON.parseToObject(kv.value(), Annotation.class));
           } else if (qual[0] == AppendDataPoints.APPEND_COLUMN_PREFIX){
+        	compactedKVTimestamp = Math.max(compactedKVTimestamp, kv.timestamp());
             final AppendDataPoints adp = new AppendDataPoints();
             tot_values += adp.parseKeyValue(tsdb, kv).size();
             last_append_column = new KeyValue(kv.key(), kv.family(), 
@@ -461,6 +465,7 @@ final class CompactionQueue extends ConcurrentSkipListMap<byte[], Boolean> {
           longest = kv;
         }
         ColumnDatapointIterator col = new ColumnDatapointIterator(kv);
+        compactedKVTimestamp = Math.max(compactedKVTimestamp, kv.timestamp());
         if (col.hasMoreData()) {
           heap.add(col);
         }
@@ -539,7 +544,10 @@ final class CompactionQueue extends ConcurrentSkipListMap<byte[], Boolean> {
       }
 
       final KeyValue first = row.get(0);
-      return new KeyValue(first.key(), first.family(), cq, cv);
+      if(tsdb.getConfig().getBoolean("tsd.storage.use_otsdb_timestamp"))
+    	  return new KeyValue(first.key(), first.family(), cq, compactedKVTimestamp, cv);
+      else
+    	  return new KeyValue(first.key(), first.family(), cq, cv);
     }
 
     /**
--- a/src/core/RequestBuilder.java
+++ b/src/core/RequestBuilder.java
@@ -1,0 +1,33 @@
+package net.opentsdb.core;
+
+import org.hbase.async.AppendRequest;
+import org.hbase.async.HBaseRpc;
+import org.hbase.async.PutRequest;
+
+import net.opentsdb.utils.Config;
+
+public class RequestBuilder {
+	
+	public static PutRequest buildPutRequest(Config config, byte[] tableName, byte[] row, byte[] family, byte[] qualifier, byte[] value, long timestamp) {
+		
+		if(config.use_otsdb_timestamp()) 
+			if((timestamp & Const.SECOND_MASK) != 0)
+				return new PutRequest(tableName, row, family, qualifier, value, timestamp);
+			else
+				return new PutRequest(tableName, row, family, qualifier, value, timestamp * 1000);
+		else
+			return new PutRequest(tableName, row, family, qualifier, value);
+	}
+
+	public static AppendRequest buildAppendRequest(Config config, byte[] tableName, byte[] row, byte[] family, byte[] qualifier, byte[] value, long timestamp) {
+		
+		if(config.use_otsdb_timestamp())
+			if((timestamp & Const.SECOND_MASK) != 0)
+				return new AppendRequest(tableName, row, family, qualifier, value, timestamp);
+			else
+				return new AppendRequest(tableName, row, family, qualifier, value, timestamp * 1000);
+		else
+			return new AppendRequest(tableName, row, family, qualifier, value);
+	}
+
+}
--- a/src/core/TSDB.java
+++ b/src/core/TSDB.java
@@ -988,12 +988,12 @@ public final class TSDB {
         Deferred<Object> result = null;
         if (config.enable_appends()) {
           final AppendDataPoints kv = new AppendDataPoints(qualifier, value);
-          final AppendRequest point = new AppendRequest(table, row, FAMILY, 
-              AppendDataPoints.APPEND_COLUMN_QUALIFIER, kv.getBytes());
+          final AppendRequest point = RequestBuilder.buildAppendRequest(config, table, row, FAMILY, 
+              AppendDataPoints.APPEND_COLUMN_QUALIFIER, kv.getBytes(), timestamp);
           result = client.append(point);
         } else {
           scheduleForCompaction(row, (int) base_time);
-          final PutRequest point = new PutRequest(table, row, FAMILY, qualifier, value);
+          final PutRequest point = RequestBuilder.buildPutRequest(config, table, row, FAMILY, qualifier, value, timestamp);
           result = client.put(point);
         }
 
@@ -1592,8 +1592,9 @@ public final class TSDB {
   /** Puts the given value into the data table. */
   final Deferred<Object> put(final byte[] key,
                              final byte[] qualifier,
-                             final byte[] value) {
-    return client.put(new PutRequest(table, key, FAMILY, qualifier, value));
+                             final byte[] value,
+                             long timestamp) {
+    return client.put(RequestBuilder.buildPutRequest(config, table, key, FAMILY, qualifier, value, timestamp));
   }
 
   /** Deletes the given cells from the data table. */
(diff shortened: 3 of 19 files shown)


karanmehta93:
I have not yet added the code for rewriting the present data as the new data with updated cell timestamps. Since the current table supports a single version and the cell timestamp will be greater than metric timestamp (since it takes a while before the metric is sent and is dumped to HBase) in most cases, we won't be able to directly write the information back to the table.

There are multiple ways to approach this, Please suggest appropriate ideas.
1. Read a KeyValue, delete it and write the new one, Major compact after the procedure is completed.
2. Use Bulk import, create new HFiles directly and swap them out with new ones.

johann8384:
I haven't done a real close review of this, but I wanted to say that you'll probably want to develop against the next branch at the very least, and 3.0 work is already happening in the 3.0 branch. Can you rebase this work against 2.4 which is in the next branch? It *should* be pretty close to 2.3 which is in the master branch.

karanmehta93:
Thanks @johann8384 for the comment, I will rebase it with the _next_ branch soon. 
One more potential issue I have identified with MockBase implementation. For every column, a new treeMap instance is instantiated, which is a map of timestamp to value and has been ordered so that the latest timestamp always stays at the top which is similar to how HBase works.
`
column = new TreeMap<Long, byte[]>(Collections.reverseOrder());
`

Since OpenTSDB uses single version, there is no need for a map here as such. Since now we are making use of timestamps for the PutRequests, there is a possibility that it may result in multiple rows being created because of multiple timestamps. However only one of them should be technically retained (the one with higher timestamp). This can lead to some failing tests. Any thoughts on this one?

manolama:
> There are multiple ways to approach this, Please suggest appropriate ideas.
> Read a KeyValue, delete it and write the new one, Major compact after the procedure is completed.
Use Bulk import, create new HFiles directly and swap them out with new ones.

It's annoying but we'd want to go with 1 probably via a CAS for atomic safety via MR or something. A number of users won't be familiar with the process of dealing with HFiles directly. And some users (like ourselves) don't have direct access to HBase to perform those operations anway.

manolama:
> ```final byte[] stored = column == null ? null : put.timestamp() != Long.MAX_VALUE ? column.get(put.timestamp()) : column.firstEntry().getValue();```
>
> @manolama Can you suggest as to how to handle this issue?
>
> One possible solution is to change the line to
```final byte[] stored = column == null ? null : column.firstEntry().getValue();```
> With this change, all the tests from TestAnnotation and TestAnnotationRpc passes.

Good catch, that should be fine to go with the null check.

**Comment by manolama** [comment]
> Since OpenTSDB uses single version, there is no need for a map here as such. Since now we are making use of timestamps for the PutRequests, there is a possibility that it may result in multiple rows being created because of multiple timestamps. However only one of them should be technically retained (the one with higher timestamp). This can lead to some failing tests. Any thoughts on this one?

For this one I'd like to maintain the behavior in MockBase similar to HBase as we may be adding some meta code soon that will require the timestamp to be functional and make sue of versions. And with 3.0 some users may want to implement a versioned store.

File: src/core/CompactionQueue.java [hunk 1]
```diff
@@ -86,6 +86,8 @@
 
   /** If this is X then we'll flush X times faster than we really need.  */
   private final int flush_speed;  // multiplicative factor
+  
+  private long compactedKVTimestamp;
```
manolama: Can't have it here as the compaction class is called from multiple threads. This works for just the scheduled compaction but the ``compact()`` method is called asynchronously.

src/core/CompactionQueue.java
```diff
@@ -539,7 +544,10 @@ private KeyValue buildCompactedColumn(ByteBufferList compacted_qual,
       }
 
       final KeyValue first = row.get(0);
-      return new KeyValue(first.key(), first.family(), cq, cv);
+      if(tsdb.getConfig().getBoolean("tsd.storage.use_otsdb_timestamp"))
+    	  return new KeyValue(first.key(), first.family(), cq, compactedKVTimestamp, cv);
+      else
+    	  return new KeyValue(first.key(), first.family(), cq, cv);
```
manolama:
```
if (tsdb.getConfig().getBoolean("tsd.storage.use_otsdb_timestamp")) {
  return new KeyValue(first.key(), first.family(), cq, compactedKVTimestamp, cv);
} else {
  return new KeyValue(first.key(), first.family(), cq, cv);
}
```
style nit pick.

src/core/CompactionQueue.java
```diff
@@ -539,7 +544,10 @@ private KeyValue buildCompactedColumn(ByteBufferList compacted_qual,
       }
 
       final KeyValue first = row.get(0);
-      return new KeyValue(first.key(), first.family(), cq, cv);
+      if(tsdb.getConfig().getBoolean("tsd.storage.use_otsdb_timestamp"))
```
manolama: This should be cached somewhere to avoid the duplicate calls to lookup and deserialize the value.

src/core/RequestBuilder.java
```diff
@@ -0,0 +1,33 @@
+package net.opentsdb.core;
```
manolama: License header.

src/core/CompactionQueue.java
manolama: Looks like you use in in L13 below :)

src/core/TSDB.java
```diff
@@ -988,12 +988,12 @@ public WritableDataPoints newBatch(String metric, Map<String, String> tags) {
         Deferred<Object> result = null;
         if (config.enable_appends()) {
           final AppendDataPoints kv = new AppendDataPoints(qualifier, value);
-          final AppendRequest point = new AppendRequest(table, row, FAMILY, 
-              AppendDataPoints.APPEND_COLUMN_QUALIFIER, kv.getBytes());
+          final AppendRequest point = RequestBuilder.buildAppendRequest(config, table, row, FAMILY, 
+              AppendDataPoints.APPEND_COLUMN_QUALIFIER, kv.getBytes(), timestamp);
```
manolama: Need to confirm timestamp is in millis here. It can come in as seconds.

manolama:
Looks good, just a few comments:

1) Check to convert the timestamps to millis for HBase to be happy
2) Thread safety in CompactionQueue
3) Style conformation with 2 space indents, brackets, etc :)

Thanks for the hard work!

karanmehta93:
Another potential issue with this patch that we should decide as to how it should be handled. 

If we write a integer value followed by a float value on the same metric timestamp, currently OpenTSDB would filter out the earlier entries of data by using the cell timestamps. Although we only maintain 1 version per timestamp, but column qualifiers with a float and integer values will be different, even with the same time offsets. This process is carried out during compaction.

After the patch, since we write the metric timestamp as cell timestamp, writes done for the same timestamp will have the same cell timestamp but will have different column qualifiers. Hence there is no way to determine the correct value of metric at that point since we cannot differentiate between them. 

1. Can anyone suggest ideas for handling this scenario?
2. How often are we to encounter such cases in practical cases? Is it fine if we can eliminate them or leave up to the code to return any value? 
Currently we have support for selecting either the oldest or the latest value, which is configurable.

I feel that this patch would require doing changes in the way OpenTSDB stores data in HBase.

src/core/CompactionQueue.java
karanmehta93: Will make the variable local to make it thread safe.

src/core/RequestBuilder.java
karanmehta93: Will add it.

src/core/TSDB.java
karanmehta93: This is handled by RequestBuilder.java, which checks the unit of timestamp via masking it and then issuing appropriate Request. If the given timestamp is in seconds. it is multiplied by 1000 to keep HBase happy!

src/core/CompactionQueue.java
karanmehta93: Will stick to the code style.

manolama:
> 1. Can anyone suggest ideas for handling this scenario?

The easiest solution here would be to switch to a value comparator when the date tiered code is enabled. Along with the data tiered setting, add a config flag to pick the largest or smallest value. Eventually we could average them :)

> 2. How often are we to encounter such cases in practical cases? Is it fine if we can eliminate them or leave up to the code to return any value?
> Currently we have support for selecting either the oldest or the latest value, which is configurable.

It doesn't happen too often but we should still make it deterministic and use the largest or smallest.

karanmehta93:
I am planning to add ValueComparator inside ColumnDatapointIterator class. Based on the configuration it will select highest or lowest value. Every value is converted to a double for comparison with the following function. Then multiple DataPoints are compared with this value. How does this approach seem?

`
public double getDoubleValue(ColumnDatapointIterator col) {

      double thisValue;
      byte[] copy = col.getCopyOfCurrentValue();
      if(Internal.isFloat(col.qualifier)) {
          thisValue = copy.length == 4 ? ByteBuffer.wrap(copy).getFloat() : ByteBuffer.wrap(copy).getDouble(); 
      } else {
          thisValue = ((copy.length == 1) ? ByteBuffer.wrap(copy).get() : ((copy.length == 2) ? ByteBuffer.wrap(copy).getShort() : ((copy.length == 4) ? ByteBuffer.wrap(copy).getInt() : ByteBuffer.wrap(copy).getLong())));
      }
      return thisValue;
  }
`

I am also not sure how we can eventually average them, since the compact method will delete all the current cells and write only the new data.

karanmehta93:
A few other potential issues found. Please suggest as to how these can be handled.
1. HBase AppendRequest doesn't take timestamp as parameter. It uses the default of HConstants.LATEST_TIMESTAMP, which boils down to EnvironmentEdgeManager.currentTimeMillis(). Since we cannot set the timestamp of the cell in HBase on such requests, it is difficult to guarantee that the cell timestamp will match the metric timestamp. DTCS can only be helpful if tsd.storage.enable_appends is set to false. The good thing is that the default option is false for it.
2. There is a deviation between the functionality of MockBase AppendRequest method and actual HBase method. If an appendRequest has a timestamp associated with it, we use the timestamp to find the exact version of the data in the map of {timestamp -> value} and try to append the given new data. However, HBase ignores timestamp and always updates the newest possible row if there are multiple versions. For our case, since we are restricted to one version, the appendRequest should update that KeyValue every time irrespective of the timestamp provided by the user. HBase automatically assigns a new timestamp to this KeyValue, which is maximum of currentTime and the cellTimestamp + 1.

manolama:
Heya @karanmehta93 Would it be better for the dedupe code to be added in https://github.com/OpenTSDB/opentsdb/blob/master/src/core/CompactionQueue.java#L479-L514 where we're already iterating for dupes?

manolama:
> 1. HBase AppendRequest doesn't take timestamp as parameter. It uses the default of HConstants.LATEST_TIMESTAMP, which boils down to EnvironmentEdgeManager.currentTimeMillis(). Since we cannot set the timestamp of the cell in HBase on such requests, it is difficult to guarantee that the cell timestamp will match the metric timestamp. Date Tiered Compaction can only be helpful if tsd.storage.enable_appends is set to false. The good thing is that the default option is false for it.

Yeah I'm not too worried here about appends compatible with DTC.

>2. There is a deviation between the functionality of MockBase AppendRequest method and actual HBase method. If an appendRequest has a timestamp associated with it, we use the timestamp to find the exact version of the data in the map of {timestamp -> value} and try to append the given new data. However, HBase ignores timestamp and always updates the newest possible row if there are multiple versions. For our case, since we are restricted to one version, the AppendRequest should update that KeyValue every time irrespective of the timestamp provided by the user. HBase automatically assigns a new timestamp to this KeyValue, which is maximum of currentTime and the cellTimestamp + 1.

Mind patching that too ;)? It'd be a pain but would be nice to have. (Unless we can actually run a real region server in memory without any other dependencies)

karanmehta93:
Thanks @manolama for your inputs.

> Would it be better for the dedupe code to be added in https://github.com/OpenTSDB/opentsdb/blob/master/src/core/CompactionQueue.java#L479-L514 where we're already iterating for dupes?

Let me look into that and will post here.

> Yeah I'm not too worried here about appends compatible with DTC.

Great. Then I will add error handling if a user tries DTC along with Append Requests. We can either throw an exception if you want or log a warning message and proceed with the regular behavior.

> Mind patching that too ;)? It'd be a pain but would be nice to have. (Unless we can actually run a real region server in memory without any other dependencies)

Yes, I will anyways have to do that for the current tests to pass. It is always better to keep our Mock as much closer to actual one as possible for future use cases as well.

karanmehta93:
> Would it be better for the dedupe code to be added in https://github.com/OpenTSDB/opentsdb/blob/master/src/core/CompactionQueue.java#L479-L514 where we're already iterating for dupes?

We are traversing over the dupes over here, but we assume that the dupes are sorted by timestamp and we pick the first one by default and discard the other ones. We iterate over them to make note of various counters we have. Since, I want to keep this logic same as before, I decided to sort them by values when I am adding them to the heap directly. 
Otherwise we would have to keep track of index offsets in the compacted_qual and compacted_val manually, depending on the config provided by the user. Please suggest what approach is better.

karanmehta93:
After resolving the current issues, I have created a new PR #990. Please have a look and suggest your inputs.

manolama:
Merged the re-work in 2112d7ed6abace19600854d31c960c6708a25955.

PR #972, opened 2017-04-22
Upgraded asyncbigtable library to v0.3.0

csoulios:
Hi, I have released asyncbigtable-0.3.0 that includes dependency upgrades. 

The most important is google-bigtable-client that removes dependencies to alpn-boot replacing it with netty-tcnative-boringssl-static. Also, it contains great performance enhancements


manolama:
Awesome, thanks! I'll see if if my old Bigtable cluster is still up to try verifying before merging. If not I'll just push it. See you at Percona tomorrow!

csoulios:
@manolama There's a small change in the bigtable configuration options

`google.bigtable.cluster.name` should be replaced with `google.bigtable.instance.id`

and `com.google.cloud.bigtable.hbase1_0.BigtableConnection` should be replaced with `com.google.cloud.bigtable.hbase1_2.BigtableConnection`

I will submit a pull request with the documentation update.

manolama:
Verified with my Bigtable instance (thanks @mbrukman!) and committed in 2316752ba6a20bd8519dbffee681aecf1f72d6a4. I'll update the docs. We also have to change ``google.bigtable.zone.name`` to ``google.bigtable.zone.id``

Issue #979, opened 2017-04-28
Query stats not written when channel is closed. 


jsbali:
Query stats don't get written in case of closed channel exception as stats.markSendFailed is never called.

jsbali:
Going through the code the fix is pretty simple (suggestions are welcome).
Fix can be in
https://github.com/OpenTSDB/opentsdb/blob/master/src/tsd/AbstractHttpQuery.java#L413 before doing a quick return call stats.markSendFailed
Also if the channel closes after that we can do that in SendSuccess onComplete and check for isSuccess and then call apt method.

johann8384:
PR welcome :)

manolama:
Merged, thanks!

PR #981, opened 2017-05-02
Log query stats in case of Closed Channel also

jsbali:
PR for https://github.com/OpenTSDB/opentsdb/issues/979
We need to change the SuccessHandler also because channel might have been closed after the 
isConnectedCheck we made.


manolama:
LGTM, thanks! Merged in bf3e374fe93673fec2cc9adda99c5f4e43152952.

PR #990, opened 2017-05-17
Fix #962 Allow OpenTSDB to take advantage of DateTieredCompaction

### LINKED ISSUE #962: Allow opentsdb to take advantage of DateTieredCompaction
DateTieredCompaction uses the scan's TimeRange to exclude store files for scanning.  This is very helpful when doing reads from timeseries data.  The problem is that it looks to be that opentsdb uses the key range to specify the query (which is great), but can we also pass along a timerange as well?  This would allow us to use this compaction strategy with opentsdb.   I assume the user can specify an Error, which is how long after we send the request that it gets persisted to hbase.  That way our TimeRange query will include a buffer of this error on both sides of our time range scan.  I can put up a patch for this if those are interested.

karanmehta93:
Following changes are done
1. RequestBuilder class added to create new PutRequest based on user parameters and configurations.
2. MockHBase functionality modified to make it equivalent to HBase.
3. Refactored and added new tests to verify the functionality
4. compactedKVTimestamp variable is now thread safe.
5. New function for merging various data points when DTCS is enabled. More details added as comments inside the code.
6. Query Scanners to use timerange based scans.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -193,7 +193,8 @@ tsdb_SRC := \
 	src/utils/JSONException.java	\
 	src/utils/Pair.java	\
 	src/utils/PluginLoader.java	\
-	src/utils/Threads.java
+	src/utils/Threads.java \
+	src/core/RequestBuilder.java
 
 tsdb_DEPS = \
 	$(COMMONS_LOGGING)	\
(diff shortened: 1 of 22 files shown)


karanmehta93:
@manolama 
Please review this PR. I have squashed all the commits and additional changes into this one. Thank you!

manolama:
Merged in 2112d7ed6abace19600854d31c960c6708a25955. Thanks!

PR #1003, opened 2017-06-26
Tags via custom HTTP header with Unit Test - provide some authentication

bhourlier:
Hi there,
This pull request add custom http header like #958 and add Unit test demand here :  https://github.com/OpenTSDB/opentsdb/pull/958#issuecomment-301331088


bhourlier:
I don't understand why my code is not validate by travis.
In the last push, all my modifications are delete and the project are similar to #958 but this project are validate by travis.
It result from modifications of test in travis ?

johann8384:
@bhourlier I'm sorry for the delay here! My daily work used to involve OpenTSDB maintenance but that was put aside for a little while. I'll take a look at this and #958 in the next few days.

PR #1031, opened 2017-07-26
Add the TimeBasedRoutingExecutor that's the first stab at a time based

manolama:
router that allows TSDB to query a cache (e.g. Facebook's Beringei)
for short-term data or fall back to long-term storage (HBase) at
query time.


manolama:
Merged in 16132f11613ecd44e1188b0467de272e34358277.

PR #1048, opened 2017-08-23
HighestCurrent may return wrong result

opsun:
Because of DataPoints maybe empty and AggregationIterator.hasNextValue() will skip empty item.


opsun:
CI failed, seem as environment issue.

manolama:
Merged in 58ab43571c1ed823102f6344a4a1429fdaef43d8, thanks!

PR #1053, opened 2017-08-31
Add the MockDataStore that holds a generated # of time series for tes…

manolama:
…ting

out the API and call graphs.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/execution/StorageQueryExecutor.java
+++ b/core/src/main/java/net/opentsdb/query/execution/StorageQueryExecutor.java
@@ -1,0 +1,166 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2017  The OpenTSDB Authors.
+//
+// This program is free software: you can redistribute it and/or modify it
+// under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 2.1 of the License, or (at your
+// option) any later version.  This program is distributed in the hope that it
+// will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+// of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser
+// General Public License for more details.  You should have received a copy
+// of the GNU Lesser General Public License along with this program.  If not,
+// see <http://www.gnu.org/licenses/>.
+package net.opentsdb.query.execution;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.google.common.base.Objects;
+import com.google.common.base.Strings;
+import com.google.common.collect.ComparisonChain;
+import com.google.common.collect.Ordering;
+import com.google.common.hash.HashCode;
+import com.stumbleupon.async.Callback;
+
+import io.opentracing.Span;
+import net.opentsdb.core.Const;
+import net.opentsdb.query.context.QueryContext;
+import net.opentsdb.query.execution.graph.ExecutionGraphNode;
+import net.opentsdb.query.pojo.TimeSeriesQuery;
+import net.opentsdb.storage.TimeSeriesDataStore;
+
+/**
+ * An executor that calls a Storage plugin to execute the query. Simple 
+ * pass-through implementation that tracks the outstanding executions.
+ * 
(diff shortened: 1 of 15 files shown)


core/src/main/java/net/opentsdb/query/execution/StorageQueryExecutor.java
```diff
@@ -0,0 +1,167 @@
...
+          && Objects.equal(storage_id, config.storage_id);
+    }
+    
+    @Override
+    public int hashCode() {
+      return buildHashCode().asInt();
+    }
+
+    @Override
+    public HashCode buildHashCode() {
```
mansu: Please document why these specific fields were chosen for hashing, equals etc..

core/src/main/java/net/opentsdb/storage/MockDataStore.java
```diff
@@ -0,0 +1,344 @@
...
+              database.entrySet()) {
+          for (Metric m : query.getMetrics()) {
+            if (Bytes.memcmp(m.getMetric().getBytes(Const.UTF8_CHARSET), 
+                             entry.getKey().metrics().get(0)) != 0) {
+              continue;
+            }
+            
+            if (!Strings.isNullOrEmpty(m.getFilter())) {
+              Filter f = null;
+              for (Filter filter : query.getFilters()) {
```
mansu: Using streams here should simplify the code and remove the null check.

core/src/test/java/net/opentsdb/storage/TestMockDataStore.java
```diff
@@ -0,0 +1,212 @@
...
+    IteratorGroups results = ex.deferred().join();
+    assertEquals(8, results.flattenedIterators().size());
+    
+    IteratorGroup group = results.group(new SimpleStringGroupId("m1"));
+    List<TimeSeriesIterator<?>> iterators = group.flattenedIterators();
+    assertEquals(4, iterators.size());
+    for (final TimeSeriesIterator<?> it : iterators) {
+      TimeSeriesIterator<NumericType> iterator = (TimeSeriesIterator<NumericType>) it;
+      long ts = 1483228800000L;
+      while (iterator.status() == IteratorStatus.HAS_DATA) {
```
mansu: I wish we can replace IteratorStatus with a Java iterator

core/src/main/java/net/opentsdb/query/execution/StorageQueryExecutor.java
```diff
@@ -0,0 +1,167 @@
...
+  private TimeSeriesDataStore data_store;
+  
+  /**
+   * Default ctor.
+   * @param node A non-null node to pull the ID and config from.
+   * @throws IllegalArgumentException if a config param was invalid.
+   */
+  public StorageQueryExecutor(final ExecutionGraphNode node) {
+    super(node);
+    if (node.getDefaultConfig() == null) {
```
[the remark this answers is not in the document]
manolama: Good idea, since I can't really think of a use case where an executor wouldn't have a default config.

core/src/test/java/net/opentsdb/storage/TestMockDataStore.java
manolama: I will shortly :)

core/src/main/java/net/opentsdb/storage/MockDataStore.java
manolama: yeah I may switch to that since this isn't a critical path, just testing.

core/src/main/java/net/opentsdb/query/execution/StorageQueryExecutor.java
manolama: Doc'd in the interface. I think.

core/src/main/java/net/opentsdb/query/execution/StorageQueryExecutor.java
```diff
@@ -0,0 +1,167 @@
...
+      throw new IllegalArgumentException("Default config cannot be null.");
+    }
+    if (node.graph() == null) {
+      throw new IllegalStateException("Execution graph cannot be null.");
+    }
+    if (Strings.isNullOrEmpty(((Config) node.getDefaultConfig()).getStorageId())) {
+      throw new IllegalArgumentException("Default storage ID must be non-empty.");
+    }
+    data_store = (TimeSeriesDataStore) node.graph().tsdb().getRegistry()
```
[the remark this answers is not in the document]
manolama: wilco.

manolama:
cleaned and merged in 03581922fe5197b576d5c0eaf490151aa2e8b112.

PR #1056, opened 2017-09-01
Add some spans to the QueryRrpc class to see what takes so long. Pars…

manolama:
…ing and

conversion is really slow on initial startup then fairly quick after that.
Also fix the completion span.


manolama:
Merged in fc71efc31467422ff935b5dba9404a9db0ad7692

PR #1076, opened 2017-09-29
Move the TimeSeriesDataType and TimeStamp interfaces to the Common mo…

manolama:
…dule.

Remove the byte TimeSeriesId and create a String based TimeSeriesId interface as
it's much more performant than the byte interface in testing. And encoded IDs
for the present TSD interface can be encoded as hex strings.
Create a new TimeSeriesValue interface in Common and simplify it so that we use
less heap, e.g. it no longer has references to the ID or a "real" value. We'll use
summary types for the latter and the TimeSeries interface will have the ID.
Introduce the TimeSpecification interface in Common for saving space and serdes
on downsampled queries.
Introduce the TimeSeries interface that contains info for all data types
associated with an ID.


common/src/main/java/net/opentsdb/data/TimeSeries.java
```diff
@@ -0,0 +1,85 @@
...
+ * {@link TimeSeriesValue} objects on each call to {@link Iterator#next()}. Each
+ * call to one of these iterator methods will return a new view of the data 
+ * starting at the beginning of the set.
+ * <p>
+ * All data types will contain data within the same time range for this time
+ * series.
+ * 
+ * @since 3.0
+ */
+public interface TimeSeries {
```
mansu: Why do we use a TypeToken instead of a generic Type?

Reflection is generally expensive but I wonder if TypeTokens have a performance hit.

Also, why not extend the Java Iterator interface? How is this interface better?

Would be awesome to document the design choices here.

common/src/main/java/net/opentsdb/data/TimeSeries.java
```diff
@@ -0,0 +1,85 @@
...
+  public TimeSeriesId id();
+  
+  /**
+   * Returns a new iterator for the type of time series data denoted by {code type}.
+   * @param type A non-null {@link TimeSeriesDataType} type of data.
+   * @return A new iterator for the type of data or null if no such type is
+   * present for the time series.
+   */
+  public Iterator<TimeSeriesValue<? extends TimeSeriesDataType>> 
+    iterator(final TypeToken<?> type);
```
mansu: Can we return an Optional<Iterator<>>? That clarifies the intent via type.

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
```diff
@@ -0,0 +1,122 @@
...
+   * A simple string for identifying the time series. The alias may be null or
+   * empty. If a value is present, the alias must be unique within a set of
+   * time series.
+   * <p>
+   * Invariant: If all other fields for the identity are empty, then the alias
+   * must be non-null and non-empty.
+   * 
+   * @return A string or null if not used.
+   */
+  public String alias();
```
mansu: Please document why this field is needed and how it's used. Also, id() may be a more consistent name than alias.

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
```diff
@@ -0,0 +1,122 @@
...
+  
+  /**
+   * A list of namespace identifiers included in the sources for this time series.
+   * May be empty if namespaces are not in use for the platform.
+   *  
+   * @return A non-null list if zero or more namespace identifiers.
+   */
+  public List<String> namespaces();
```
mansu: What is a namespace? A link to documentation may help here.

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
```diff
@@ -0,0 +1,122 @@
...
+   * A list of metric names included in the sources for this time series.
+   * If this list is empty, check {@link #metrics()} for data.
+   * <p>
+   * Invariant: If this list is empty (meaning the series or set of series is 
+   * not associated with a specific metric) then the {@link #alias()} must be 
+   * non-null and unique.
+   *  
+   * @return A non-null list of zero or more metric names.
+   */
+  public List<String> metrics();
```
mansu: Rename this method metricNames?

Also, why would a time series have multiple metric names?

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
```diff
@@ -0,0 +1,122 @@
...
+   * - The underlying store does not support tags.
+   * - All of the source time series shared the same tag pairs.
+   * - All of the source time series had disjoint tag sets.
+   * <p>
+   * Invariant: Each tag name must be present in every time series if it appears
+   * in this list.
+   * 
+   * @return A non-null list of zero or more tag names.
+   */
+  public List<String> aggregatedTags();
```
mansu: I think it would be a good idea to have one interface for data and one interface for the response containing aggregated tags etc.. That way we can use the basic data interface in a store like metron.

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
```diff
@@ -0,0 +1,122 @@
...
+  public Map<String, String> tags();
+
+  /**
+   * A list of tag names (tagk) that were represented in every source series
+   * but had one or more differing tag values (tagv). This list may be empty if
+   * - The underlying store does not support tags.
+   * - All of the source time series shared the same tag pairs.
+   * - All of the source time series had disjoint tag sets.
+   * <p>
+   * Invariant: Each tag name must be present in every time series if it appears
```
mansu: Can we also document the purpose of aggregated tags. It is unclear how these are useful.

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
```diff
@@ -0,0 +1,122 @@
...
+   * - All of the source time series shared the same tag pairs.
+   * - All of the source time series shared the same tag names with different
+   * values.
+   * <p>
+   * Invariant: Each tag name must be present in at least one but not all 
+   * source series if it appears in this list.
+   * 
+   * @return A non-null list of zero or more tag names.
+   */
+  public List<String> disjointTags();
```
mansu: Can we please document how disjoint tags are useful.

common/src/main/java/net/opentsdb/data/TimeSeriesValue.java
```diff
@@ -20,17 +20,13 @@
...
  * <p>
  * <b>WARNING:</b> Note that when a value is extracted from an iterator, a copy
- * should be made using {@link #getCopy()}.
+ * should be made as the iterator may change the actual value on the next
+ * iteration.
  *
  * @param <T> A {@link TimeSeriesDataType} object.
  * @since 3.0
  */
 public interface TimeSeriesValue<T extends TimeSeriesDataType> {
```
mansu: Looks great!

common/src/main/java/net/opentsdb/data/TimeSpecification.java
```diff
@@ -0,0 +1,70 @@
...
+   * The start timestamp of the interval bound by this specification, inclusive.
+   * @return A non-null timestamp object.
+   */
+  public TimeStamp start();
+  
+  /**
+   * The end timestamp of the interval bound by this specification, inclusive.
+   * @return A non-null timestamp object.
+   */
+  public TimeStamp end();
```
mansu: To clarify the non-null intent we may add a @NotNull annotation to the methods and fields. The compiler will ignore these annotations but the intent is captured better.

common/src/main/java/net/opentsdb/data/TimeSpecification.java
```diff
@@ -0,0 +1,70 @@
...
+  
+  /**
+   * The interval between timestamps in {@link #units()}. Must be a value greater
+   * than zero.
+   * @return The interval between timestamps.
+   */
+  public long interval();
```
mansu: Can we return a Duration instead?
https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html

common/src/main/java/net/opentsdb/data/TimeSpecification.java
```diff
@@ -0,0 +1,70 @@
...
+  
+  /**
+   * A helper that updates the given timestamp with a new time based on the
+   * interval offset from the start time. E.g. an offset of zero would set the
+   * time at {@link #start()}. An offset of 1 would set the time at {@link #start()}
+   * + {@link #interval()}.
+   * @param offset A zero or greater interval offset.
+   * @param timestamp A non-null timestamp to update with the new time.
+   */
+  public void updateTimestamp(final int offset, final TimeStamp timestamp);
```
mansu: I think we can use Duration here as well.

core/src/main/java/net/opentsdb/data/types/numeric/NumericType.java
```diff
@@ -64,21 +64,21 @@
    * @return {@code true} if the value is of integer type (call {@link #longValue()}), 
    * {@code false} if it's of double point type (call {@link #doubleValue()}.
    */
-  public abstract boolean isInteger();
+  public boolean isInteger();
```
mansu: With TypeTokens do we still need type checks like these?

core/src/main/java/net/opentsdb/data/types/numeric/UglyByteNumericSerdes.java
```diff
@@ -60,43 +60,50 @@ public void serialize(final TimeSeriesQuery query,
     try {
       final TimeSeriesId id = data.id();
       
-      stream.write(Bytes.fromInt(id.alias() == null ? 0 : id.alias().length));
-      if (id.alias() != null && id.alias().length > 0) {
-        stream.write(id.alias());
+      byte[] buf = id.alias() == null ? null : id.alias().getBytes(Const.UTF8_CHARSET);
```
mansu: Pick a better name for the class?

Also, can we replace this encoding with gorilla encoding?

core/src/main/java/net/opentsdb/storage/MockDataStore.java
```diff
@@ -80,19 +78,20 @@
...
       type = Maps.newHashMap();
-      database.put(value.id(), type);
+      database.put(id, type);
     }
     
     MockSpan<?> span = type.get(value.type());
     if (span == null) {
       if (value.type() == NumericType.TYPE) {
-        span = new MockSpan<NumericType>();
+        span = new MockSpan<NumericType>(id);
```
mansu: Pass in an Optional<Span>?

common/src/main/java/net/opentsdb/data/TimeSeries.java
manolama: > Why do we use a TypeToken instead of a generic Type?
>Reflection is generally expensive but I wonder if TypeTokens have a performance hit.

I like the query methods available with TypeToken and it should only get into reflection if we use some of the more advanced queries. For simple equality, the bulk of the ops here, it shouldn't incur a hit.

> Also, why not extend the Java Iterator interface? How is this interface better?

This is the higher-level collection of iterators for the time series, going back to returning iterators for numbers, strings, etc associated with a time series. This will return plain old java iterators.

> Would be awesome to document the design choices here. 
I will :)

common/src/main/java/net/opentsdb/data/TimeSeries.java
manolama: Good idea.

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
manolama: Yeah, I'll doc all these. This is used for variable naming in query expressions.

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
manolama: Will link. Used for multi-tenancy

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
manolama: For query operations wherein we have multiple metrics. Depending on the operation these can accumulate metric names from multiple series or just be replaced.

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
manolama: aye

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
manolama: Hmm, We could do that. This one has a lot of bits for the query ID vs the raw data ID.

common/src/main/java/net/opentsdb/data/TimeSeriesId.java
manolama: aye

common/src/main/java/net/opentsdb/data/TimeSpecification.java
manolama: good point.

common/src/main/java/net/opentsdb/data/TimeSpecification.java
manolama: Sure. I'm a little worried about irregular durations but we'll try it.

common/src/main/java/net/opentsdb/data/TimeSpecification.java
manolama: This takes the built in `Duration` and multiplies it by `offset`. I don't want to offer a duration here as users could break the spec by passing in irregular durations.

core/src/main/java/net/opentsdb/data/types/numeric/NumericType.java
manolama: Yeah, for the numeric since I want to use the same interface for int/floats. The type tokens are for the higher level numerics vs annotations.

core/src/main/java/net/opentsdb/data/types/numeric/UglyByteNumericSerdes.java
manolama: heh, yeah we definitely can and should replace this with gorilla encoding. It's just a sample class.

core/src/main/java/net/opentsdb/storage/MockDataStore.java
manolama: ID moved to a higher level to avoid duplication so it's no longer used here.

manolama:
Merged in f9555a49620162b204d2bec33b1d10c0ad8d992e, thanks.

PR #1077, opened 2017-09-29
Adding QueryLimitOverride to the list of build files

hkousha:
### Change:
QueryLimitOverride.java seems to be missed out from the Makefile, causing the build to fail.

### Test:
 - Created an rpm package


manolama:
Merged in 607f7d89f78ac75620404e827f42d0612ff4b28a, thanks!

PR #1084, opened 2017-10-13
Finally change the ID so that it has one namespace and one metric. We…

manolama:
… can shift the

combos into the uniques.


manolama:
Merged in c72434e259a6fe2cc1689b4ab07b4e8f5a1488f9, thanks

PR #1085, opened 2017-10-14
Add more query interfaces for query nodes, results contexts and pipel…

manolama:
…ines.

Also add a stats package with interfaces for query stats and tracers.
Include a mock tracer implementation for unit testing and export the test
package for use in other modules.


common/src/main/java/net/opentsdb/query/QueryNodeConfig.java
```diff
@@ -0,0 +1,27 @@
...
+// see <http://www.gnu.org/licenses/>.
+package net.opentsdb.query;
+
+/**
+ * The configuration interface for a particular query node. Queries will populate
+ * the configs when instantiating a DAG.
+ * 
+ * @since 3.0
+ */
+public interface QueryNodeConfig {
```
sidhhu: Will you be adding more methods to this interface?

manolama: Possibly, for now it's just a placeholder with the ID.

manolama:
Merged in e14384f4982e35a3634d13e58648b6c5b0d1a892

PR #1092, opened 2017-10-30
** COMMON **

manolama:
Create the TSDB and Registry interfaces.
Now TSDB.java is DefaultTSDB.java implementing the TSDB interface and
the registry implementation will be renamed.
Convert the TSDBPlugin to an interface with an abstract implementation.
Add the Aggregator interface. Old TSDB aggregators must implement this.
Add the TimeSeriesDataSource that moves fetchNext() into an interface
that only sources implement. We can't fetchNext() on an intermediate
node.
Add the QueryContextBuilder interface to build a query
Add a QueryFillPolicy interface with a new enum that allows fills to
choose how to handle real values, e.g. fill with real next or previous values
when available.
Add the QueryIterator interface that allows nodes to generate synthetic
time series.
Add the QueryIteratorFactory to return iterators from a QueryNode at run-time.
Add QueryIteratorInterpolator that is an interface for interpolating values
not present at a specific time stamp for a time series (and a config interface)
Add QueryIteratorInterpolatorFactory for generating the interpolators.

** CORE **
Move SlicedTimeSeriesIterator to SlicedTimeSeries
Move core/Aggregators to date/types/numeric/Aggregators and implement the
new aggregator interface.
Add BaseNumericAggregator as an abstract for numeric aggregators.
Add BaseNumericFillPolicy to handle the common fill policies and serve as
a base for other types of fills.
Add ctor and reset to MutableNumericType that can pull from NumericType values.
Add NumericAggregator as an interface that numeric aggregators must implement.
Add NumericInterpolatorFactories to hold implementations of fill policies and
interpolators. Needs some work.
Add ScalarNumericFillPolicy to fill with fixed numbers.
Add NumericInterpolator as an implementation of the QueryIteratorInterpolator
as the base interpolator.
Add NumericLERP to implement the current default linear interpolation.
Add the GroupBy processor and sub classes to aggregate time series into one
value.


manolama:
Merged in bcdf397e7c6d00a5675da8f40d99d7a5e300a77e

PR #1093, opened 2017-11-01
Return the query via QueryContext interface.

manolama:
Force QueryNodeFactories to implement the TSDBPlugin interface.
Return the Span from the common Span set methods and add the log() method for
exceptions as well as a newChild() method so users can easily create a child
span from an existing span.
Create the common Trace interface that tracks a specific tracing instance.
Change the Tracer interface to return Traces and implement the TSDBPlugin
interface.
Fix some classes referring to DefaultTSDB.
Implement caching in the getQueryNodeFactory() in DefaultRegistry().
Disable the execution graph loading by default in DefaultRegistry() for now.
Change TimeSeriesSerdes so that it doesn't require a type and takes a
QueryContext and QueryResult as parameters.
Modify JsonV2QuerySerdes.java to work with the new QueryResult objects.
Modify QueryRpc to execute a the new query style against the QueryContext and
QueryResult. For now it ONLY works with the MockDataStore.
Implement the new tracing APIs with the Brave tracing plugin.

### DIFF
--- a/common/src/main/java/net/opentsdb/query/QueryContext.java
+++ b/common/src/main/java/net/opentsdb/query/QueryContext.java
@@ -54,4 +54,7 @@ public interface QueryContext {
    * @return An optional stats collector for the query, may be null.
    */
   public QueryStats stats();
+  
+  /** @return The original query. */
+  public TimeSeriesQuery query();
 }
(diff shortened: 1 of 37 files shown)


manolama:
Merged in 057de0e4571e9a8872ef77029592db7e700511b5

PR #1100, opened 2017-11-08
COMMON:

manolama:
Change the UTC timezone in Const to a ZoneId to use the new Java time classes.
Move MillisecondTimeStamp to Common
Add TimeSeriesValueComparator to Common to sort and/or compare time series
values based on their timestamps.
Change the interval to a TemporalAmount instead of Duration in the
TimeSpecification class and add the timezone interface.
Add nextTimestamp() to the TimeSpecification class to increment the timestamp.
Rename TimeStamp's TimeStampComparator to RelationalOperator to make it more
obvious and differentiate from the new comparator.
Add a nanos() getter to the TimeStamp class so we can now support nanosecond
precision.
Add a timezone getter to TimeStamp so it can store a time zone for
alignment purposes.
Add a reset method with epoch and nanos to the TimeStamp.
Provide the add() interface for the TimeStamp to change the value by a
TemporalAmount.
Add snapToPreviousInterval() methods to the TimeStamp class so that the values
can be aligned on calendar boundaries when downsampling. This allows for
different implementations and takes the alignment out of the DateTime class.
Add a new TimeStampComparator that implements the Java Comparator interface for
comparing timestamp orders.
Add the ZonedNanoTimeStamp implementation that allows for nanosecond precision
and snapping to time-zone based calendars.
Add a MockTimeSeries class that allows processors to easily mock out data.
Add unitsToChronoUnit() DateTime class and remove the old calendar based
method.
Add ParseDuration2() that returns a temporalAmount.
TODO - need to fix up parseDateTimeString() for the new date objects.

CORE:
Modify MutableNumericType to allow for null time series value values (timestamp)
cannot be null though. Also add a resetNull() function.
Implement Downsample classes to perform calendar based downsampling of higher
resolution data to lower resolutions using an aggregator.
Add BaseQueryNodeFactory as a base implementation for factories.
Modify the GroupBy class to return a single result per received result and keep
track of the number of outstanding upstream consumers before closing the
original result.


manolama:
Merged in ff49537d3593d1751036367c2a72425c2f6da6db, thanks!

PR #1101, opened 2017-11-10
Add nanos and micros to the DateTime.getDurationUnits() method.

manolama:
Modify MutableNumericType to clone higher resolution timestamps when given
in a reset operation.
Modify RateOptions to support a string interval instead of an integer and
have it implement QueryNodeConfig.
Add the Rate processor for computing rates and put it in the same order as it
runs in 2.0.


manolama:
Merged in afadf614d516bdc715fcff0ba7c1893dbac81aef, thanks.

Issue #1105, opened 2017-11-14
opentsdb default encode charset and UID hash collision 


huainansto:
OPENTSDB default encode charset is ISO-8859-1， when deal with Chinese characters, will get messy code, need change default encode charset to UTF-8。
We often encounter the following cases
Caused by: java.lang.IllegalStateException: id=[-75, -68, -116, -12, -81, -39, -58] => name='X', already mapped to 'Y'
The reason is ConcurrentHashMap instance 'id_cache' hash collision in file 'UniqueId.java'

IDerr:
Is is possible to translate in english ?

Thanks :D

huainansto:
1. OPENTSDB default encode charset is ISO-8859-1， when deal with Chinese characters,  will get messy code, need change default encode charset to UTF-8。
2. We often encounter the following cases
Caused by: java.lang.IllegalStateException: id=[-75, -68, -116, -12, -81, -39, -58] => name='X', already mapped to 'Y'
The reason is ConcurrentHashMap instance 'id_cache' hash collision in file 'UniqueId.java'

IDerr:
Thanks :)

Tony86Yan:
Default charset change to utf-8, also got this problem…

manolama:
Right 3.0 supports UTF-8 now but older versions need a bunch of code changes for UTF-8 support and you have to truncate the UID and data tables when you make such a change.

PR #1107, opened 2017-11-14
修改默认编码为UTF-8

huainansto:
修复id_cache 由于编码导致的hash碰撞

### DIFF
--- a/src/core/Const.java
+++ b/src/core/Const.java
@@ -47,7 +47,7 @@ public final class Const {
    * Charset to use with our server-side row-filter.
    * We use this one because it preserves every possible byte unchanged.
    */
-  public static final Charset ASCII_CHARSET = Charset.forName("ISO-8859-1");
+  public static final Charset ASCII_CHARSET = Charset.forName("UTF-8");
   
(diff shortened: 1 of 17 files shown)


huainansto:
fix Issues #1105 [https://github.com/OpenTSDB/opentsdb/issues/1105](url)

johann8384:
Original Title: "OPENTSDB default encode charset is ISO-8859-1， when deal with Chinese characters, will get messy code, need change default encode charset to UTF-8。 We often encounter the following cases Caused by: java.lang.IllegalStateException: id=[-75, -68, -116, -12, -81, -39, -58] => name='X', already mapped to 'Y' The reason is ConcurrentHashMap instance 'id_cache' hash collision in file 'UniqueId.java'"

manolama:
I'd like to but we have to make this an option for the 2.x branch as so many folks have written data already and just switching will break it (as I found out in 2.4).

ace-han:
@manolama What are your suggestions for those using 2.x branch to deal with UTF charset before 3.x?

ace-han:
@manolama any extra config need to do? Or just fresh tables in HBase?

PR #1122, opened 2017-12-25
add_utf8_support

xianhuawei:
the opentsdb v2.3.0 is not support chinese !  
the ISO-8859-1 code is messy code !


johann8384:
This seems to be a duplicate of #1107

PR #1144, opened 2018-02-23
Start a new configuration system that allows loading and merging from

manolama:
multiple sources along with dynamic value updates and nice debug info.
Based on cfg4j and Nick's work with the fat jar.
Move ArgValueValidator and ConfigArgP to common/config.

Rename ArgValueValidator to ConfigValueValidator and have it work over
ConfigurationItems

Add the ConfigTypeRef based on Jackson's method for capturing the
type of a complex object.

Add a new Config interface that will be used for more advanced configuration
of the TSD.
Also add a ConfigCallback class to use for binding config settings.

Rename ArgValueValidator to ConfigValueValidator.
Comment out bits of the ConfigArgP class we don't need now.
Add ConfigEntrySchema with a builder that will make it easier
to programatically create these.
Add tests for ConfigEntrySchema and ConfigTypeRef.

Add the ConfigSetting class for overriding the schema defaults.

Add the ConfigEntry class that is meant to be the value in a
concurrent hash map keyed on the setting key. The class holds
the schema, overrides and callbacks.


manolama:
Moving this around a bit but yeah we'll be able to pull from ZK and Consul.

PR #1145, opened 2018-02-26
Create a new Configuration module that includes support for:

manolama:
- multiple local and remote sources flattened into a single value
- reloads of local and remote sources
- binding to keys so that reloads update configs in the app
This is based off work Nick did and cfg4j so that OpenTSDB can
pull it's config from locations other than a local file, thus
making it much easier to deploy in containerized environments.


manolama:
Merged in 1d3aa42802151dabd71f8d08eaf200fcd6f1487b

PR #1146, opened 2018-02-27
COMMON:

manolama:
- Rename the RegisterSchema() methods to Register() and add some overloads that
allow for registering primitive values to avoid having to create complex
builders anytime you need to register in code.
- Rename hasKey() to hasProperty() so it's more compatible with the old
config class.
- Add a UnitTestConfiguration class to make it easier to use the configuration
object in unit tests.
- Explicitly load the UnitTest and RuntimeOverride providers for UTs that are
using Mockito or PowerMockito.
- Add a helper to Threads to get the calling class name two methods up so that
config registrations can capture the source.
- Return the new Configuration class from the TSDB interface.

CORE:
- Use the new Configuration class.

IMPLEMENTATIONs:
- Use the new Configuration class.
- Brave tracer will allow for changing endpoints. The new Config class allows
binding so there's a callback now that will attempt to kill and start a new
reporter.
- Fix QueryRpc null trace entries in the immutable map.


manolama:
Merged in c1c186802863b935e8e681179c8bdca5438f71fa

PR #1150, opened 2018-03-01
CORE:

manolama:
- Modify the TagVLiteralOrFilter to dedupe in a set but return a list
  so we can maintain order during resolution

STORAGE:
- Implement and test UniqueId.resolveUids()


manolama:
Merged in c0d6144a049afb73c9257f3df0e92eb54a6212c1

PR #1151, opened 2018-03-02
CORE:

manolama:
- Add CacheMode to the UniqueId class to determine which cache we're
  supposed to populate.
- Have the UniqueIdStore return the character set.
- Add the LRU UniqueId cache with Guava implementation of the getId(s)
  and getName(s) implementations.


manolama:
merged in 3d0391343c7729eb9f56bca7c093a04df7109a8a

PR #1162, opened 2018-03-10
CORE:

manolama:
- remove the MAX_TIMESPAN from Const and put it in Schema.
- Move the RollupUsage enum into RollupUtils.
- Pull setBaseTime in from Internal into Schema and remove some function stubs
  we won't actually need.


manolama:
Merged in 86b1003441c0e9d1d20d4e07fa1f3f57f39ab166

PR #1165, opened 2018-03-16
CORE:

manolama:
- Schema now returns a null RollupConfig instead of throwing the exception till
  we finish the implementation.

STORAGE:
- Add a table getter to the MockScanner in MockBase
- Change the scanners set to a list in MockScanner so we can verify easier
- Tweak UTBase to return the default data table
- Change the QueryUtil methods to take a map of lists of byte arrays instead
  of a map of arrays of byte arrays.
- Change Tsdb1xHBaseDataStore to stub out dynamic query properties and
  add getters for various types like string, ints and booleans. Also add some
  static key strings.
- Add some stubs to Tsdb1xQueryNode for scanner behavior.
- Add isFull() to Tsdb1xQueryResult.
- Tweak Tsdb1xScanner to use the overrides in the node and result.

### DIFF
--- a/core/src/main/java/net/opentsdb/storage/schemas/tsdb1x/Schema.java
+++ b/core/src/main/java/net/opentsdb/storage/schemas/tsdb1x/Schema.java
@@ -557,7 +557,7 @@ public class Schema implements StorageSchema {
   
   public RollupConfig rollupConfig() {
     // TODO - implement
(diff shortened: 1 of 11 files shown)


manolama:
26a47d0ada577fba58dfb0ab41cf68b801d8ca47

PR #1166, opened 2018-03-16
3x tweaks

manolama:


storage/asynchbase/src/main/java/net/opentsdb/storage/Tsdb1xScanners.java
```diff
@@ -14,57 +14,965 @@
...
+    synchronized (this) {
+      if (current_result != null) {
+        throw new IllegalStateException("Query result must have been null "
+            + "to start another query!");
+      }
+      current_result = result;
+    }
+    
+    // just extra safe locking. Shouldn't ever happen.
+    if (!initialized) {
```
SeanPMiller: You have DCLP on `initialized` field here, but it's not marked `volatile`.

storage/asynchbase/src/main/java/net/opentsdb/storage/Tsdb1xScanners.java
```diff
@@ -14,57 +14,965 @@
...
+   * @param span An optional span.
+   * @throws IllegalArgumentException if the result is null.
+   */
+  public void fetchNext(final Tsdb1xQueryResult result, final Span span) {
+    if (result == null) {
+      throw new IllegalArgumentException("Result must be initialized");
+    }
+    
+    synchronized (this) {
+      if (current_result != null) {
```
SeanPMiller: Are you expecting to see the memory effects from other threads here? That is, could another thread set `current_result` to `null`?

storage/asynchbase/src/main/java/net/opentsdb/storage/Tsdb1xScanners.java
```diff
@@ -14,57 +14,965 @@
...
+  boolean hasException() {
+    return has_failed;
   }
   
+  /** Called by a child when the scanner has finished it's current run. */
   void scannerDone() {
-    // TODO - implement
+    boolean send_upstream = false;
+    synchronized (this) {
+      scanners_done++;
```
SeanPMiller: Likewise, is the value of this variable meant to be memory-consistent across threads?

storage/asynchbase/src/main/java/net/opentsdb/storage/Tsdb1xScanners.java
manolama: Another thread could and they need to be blocked. It would mean there was a bug upstream.

PR #1169, opened 2018-03-20
3x tweaks

manolama:


storage/asynchbase/src/main/java/net/opentsdb/storage/Tsdb1xMultiGet.java
```diff
@@ -0,0 +1,742 @@
...
+      }
+    }
+  }
+
+  /**
+   * Increments the main timestamp or the fallback timestamp depending
+   * on the {@link #rollup_index}.
+   */
+  @VisibleForTesting
+  void incrementTimestamp() {
```
sidhhu: Please add reverse scanning to this.

PR #1174, opened 2018-03-26
CORE:

manolama:
- Add schema documentation to the NumericCodec.
- Move and add methods to NumericCodec to encode/decode append columns mostly
  for unit tests but we may be able to use them in the write path. This will
  clean up the NumericType class a bit for things that should be schema
  specific.


manolama:
Merged in a82bed539d91de62c68680b649358950b3c5f395

PR #1177, opened 2018-03-28
CORE:

manolama:
- Add the Configuration to QuerySourceConfig so downstream operations
  have access. Also add validation and javadocs. And UTs.
- Add size() and dataPoints() to the RowSeq interface so we can extract that
  info.
- Add a static guestimate for the NumericRowSeq size. We could use JOL but it
  would be too expensive in the fast path.
- Add byte and dp limit keys to the Schema and register them with the config.
- Add the Tsdb1xQueryResult base class for storing row sequences in the proper
  time series.


manolama:
Merged in 6ce385a5dd32e8730dbc6bc7e1ef0e0c50081149

PR #1178, opened 2018-03-29
CORE:

manolama:
- Add newRowSeq to the Codec class and have NumericRowSeq return it.
- Add reverse and keep earliest settings to the Schema class. We may move those
  out to something more global later on.
- Add reversed and keep earliest fields to the Tsdb1xQueryResult class and
  clean out the addSequence method since they're not needed.

STORAGE:
- AsyncHBase's Tsdb1xQueryResult now extends the schema base and implements
  a more efficient decode method that will take the whole row, build row seqs
  from the results, then dedupe each sequence before adding it to the result
  map. That keeps us from sorting on every data point insertion.
- Update the other classes to use the new decode method.


manolama:
Merged in 37c53661478ee34208c5c15b0d6e469beb7eecd1

PR #1182, opened 2018-04-02
COMMON:

manolama:
- Remove StorageSchema
- Change the TimeSeriesByteId to require a TimeSeriesDataStore instead of the
  StorageSchema.
- Add the resolveByteID() method (with a span) to the TimeSeriesDataStore and
  take out the id() and version() methods that will be implemented by the
  QueryNode plugin.

CORE:
- Tweak BaseTimeSeriesByteId and MergedTimeSeriesId.javato use the
  TimeSeriesDataStore.
- Modify MockDataStore to return an UnsupportedOperationException when
  resolving IDs.
- Change TSDBV2Pipeline to get the TimeSeriesDataStore factory and perform some
  null checks.
- Add MockDataStoreFactory.
- Modify JsonV2QuerySerdes to resolve binary IDs to strings before writing
  the JSON.
- Change NumericMillisecondShard to take the base TimeSeriesId instead of the
  string version.
- Delete the StorageSchema and factory interfaces. To keep the API simpler, Schema
  just implements a TimeSeriesDataStore.
- Tweak Schema to load the default data store of the new Tsdb1xDataStoreFactory
  and have it implement the TimeSeriesDataStore instead of the StorageSchema.
  Still stubbed out.
- Don't drop the map on Tsdb1xTimeSeries close as the v2 pipeline accumulates
  the results and calls close. TODO - figure out if we should copy the series or
  what.
- Add SchemaFactory to instantiate the Schema data store.

STORAGE:
- Build a fat JAR for the AsyncHBase storage driver.
- Add the Tsdb1xHBaseFactory to instantiate HBase drivers.
- Move the Stage enum to HBaseExecutor and require the executors to return a
  a state.
- Have the Tsdb1xHBaseDataStore implement the Tsdb1xDataStore interface and
  accept a schema. Also register some configs here.
- Modify Multiget to return a state.
- Add state to Scanners
- The Tsdb1xQueryNode can't initialize itself. The graph has to. Also the
  query node will now send the onCompete() upstream if the executor is in the
  state complete.


SeanPMiller:
The changes look reasonable enough, but I want to discuss (offline, probably) the reasoning behind the *Schema changes. Simplicity seems to be the driver.

manolama:
Yup simplicity from the query standpoint, but then I also have to share it with Bigtable so it's a bit convoluted and can't be tied just to HBase.

manolama:
Merged in a55b0880b38a58c0a55685201e4a38ed87cca7f4

PR #1183, opened 2018-04-03
ALL:

manolama:
- Bump the build version.

STORAGE:
- Add AsyncHBase flags and set them in the config.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/TSDBV2Pipeline.java
+++ b/core/src/main/java/net/opentsdb/query/TSDBV2Pipeline.java
@@ -16,9 +16,14 @@ package net.opentsdb.query;
 
 import java.time.ZoneId;
 import java.util.Collection;
+import java.util.List;
+import java.util.Set;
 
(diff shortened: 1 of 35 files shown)


core/src/main/java/net/opentsdb/query/TSDBV2Pipeline.java
```diff
@@ -101,6 +106,7 @@ public void initialize() {
       Filter filter = Strings.isNullOrEmpty(metric.getFilter()) ? null : q.getFilter(metric.getFilter());
       if (filter != null) {
         GroupByConfig.Builder gb_config = null;
+        final Set<String> join_keys = Sets.newHashSet();
```
sidhhu: Does it have to be concurrent?

manolama: nope

manolama:
Merged in 1018893c3a22ccc50ba2807dc15b504d98ef6fc0

PR #1187, opened 2018-04-07
CORE:

manolama:
- Move the 2x downsampling before group by as it was previously.
- Fix a sorting bug in the NumericSpan class.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/TSDBV2Pipeline.java
+++ b/core/src/main/java/net/opentsdb/query/TSDBV2Pipeline.java
@@ -103,6 +103,28 @@ public class TSDBV2Pipeline extends AbstractQueryPipelineContext {
       QueryNode node = store.newNode(this, config);
       addVertex(node);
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/net/opentsdb/query/TSDBV2Pipeline.java
```diff
@@ -103,6 +103,28 @@ public void initialize() {
       QueryNode node = store.newNode(this, config);
       addVertex(node);
 
+      final Downsampler downsampler = metric.getDownsampler() != null ? 
+          metric.getDownsampler() : q.getTime().getDownsampler();
+      // downsample
+      if (downsampler != null) {
+        DownsampleConfig.Builder ds = DownsampleConfig.newBuilder()
+            .setId("downsample_" + metric.getId())
```
sidhhu: define a variable for "downsample_" if you will use it in other places.

manolama:
Merged in c7f6eac98e73ec4e642264750a8397b82b7c4b1e

PR #1192, opened 2018-04-24
CORE:

manolama:
- Modify the group-by and TSDBV2 pipeline so it can aggregate on all values in the
  same way the 2.x TSD does. If the user supplies a "none" aggregator we'll skip
  group by. Also add more UTs.

SERVLET:
- Add an async timeout to the query rpc handler.


manolama:
Merged in 1c4458e07a90f2bc2d0bf6b5056fcf7191672684

PR #1193, opened 2018-04-25
CORE:

manolama:
- Rework the NumericAggregator interface to include a mutable data point that
  each agg must update with a value. This saves having to capture a result and
  instantiate a NumericType value every time we agg. Instead we just use a
  stack pointer.
- Also allow for infectious NaNs in double aggregators.
- Add aggregator TODOs for long arrays where we may overflow.
- Add "multiply" and "stdev" names for the MULTIPLY and DEV aggs as alternates.
- Add resetValue() and resetTimestamp() calls to MutableNumericType along with
  a toString().
- Pass the query to TsdbV2Pipeline's downsampler and set it's fill policy
  properly.
- Tweak JsonV2QuerySerdes to handle serializing null fills.
- TODO in the DownsampleNumericIterator to add an infectious nan config.
- Fix the GroupByNumericIterator to return the proper mutable value value and
  handle nulls and NaNs from downstream.
- Remove TestDefaultTimeSeriesIterators.
- Rename MockNumericIterator MockNumericTimeSeries and clean it up.


core/src/main/java/net/opentsdb/data/types/numeric/Aggregators.java
```diff
@@ -215,21 +216,42 @@ public Sum(final String name) {
...
+      int nans = 0;
       for (int i = 0; i < limit; i++) {
+        if (Double.isNaN(values[i]) && !infectious_nans) {
+          nans++;
+          continue;
+        }
         sum += values[i];
       }
-      return new NumericValue(sum);
+      if (nans == limit) {
```
SeanPMiller: Fix condition.

SeanPMiller: Also add a test case to exercise branch.

PR #1194, opened 2018-05-07
Allow TagVNotLiteralOrFilter to filter single character tag values

neilfordyce:
Bit of an edge case.  The not_literal_or filter does not work for single character tag values.  So `zimsum:sys.cpu.nice{}{host=not_literal_or(5)}` fails, but `zimsum:sys.cpu.nice{}{host=not_literal_or(50)}` and `zimsum:sys.cpu.nice{}{host=not_literal_or(|5)}` succeed.

Fixed by borrowing the validation logic from src/query/filter/TagVLiteralOrFilter.java.


manolama:
Merged in 0c307b0a603dbbe0dd74e9412343e7faf3cc1a5b, thanks!

PR #1201, opened 2018-05-10
COMMON:

manolama:
- Add StatsCollector interface to the common lib that provides interfaces for
  collecting metrics.
- Add a blackhole implementation of the stats collector.
- Add the stats collector to the TSDB interface along with a timer getter.

CORE:
- Remove the old collectStats method from the BaseTSDBPlugin.
- Load stats plugin in DefaultTSDB and return the timer.
- Fix the ctors in a few places that had DefaulTSDB instead of TSDB.
- Use the new stats collector in the Guava LRU classes.

REDIS:
- Use the new stats collector. TODO - tags.

SERVLET:
- Just playing with stats right now.

### DIFF
--- a/implementation/servlet/src/main/java/net/opentsdb/servlet/resources/QueryRpc.java
+++ b/implementation/servlet/src/main/java/net/opentsdb/servlet/resources/QueryRpc.java
@@ -51,6 +51,7 @@ import com.google.common.collect.Sets;
 
 import jersey.repackaged.com.google.common.collect.ImmutableMap;
 import net.opentsdb.core.DefaultTSDB;
+import net.opentsdb.core.TSDB;
 import net.opentsdb.core.Tags;
 import net.opentsdb.exceptions.QueryExecutionException;
 import net.opentsdb.query.DefaultQueryContextBuilder;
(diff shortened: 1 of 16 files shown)


implementation/servlet/src/main/java/net/opentsdb/servlet/resources/QueryRpc.java
```diff
@@ -1319,12 +1338,12 @@ static final public RateOptions parseRateOptions(final boolean rate,
 //    return query;
 //  }
   
-  /** @param collector Populates the collector with statistics */
-  public void collectStats(final StatsCollector collector) {
-    collector.record("http.query.invalid_requests", query_invalid);
-    collector.record("http.query.exceptions", query_exceptions);
-    collector.record("http.query.success", query_success);
-  }
+//  /** @param collector Populates the collector with statistics */
```
SeanPMiller: Did you mean to remove this?

manolama:
merged in f418c846d3d33e1dae208f0c43257c36441cf82f

implementation/servlet/src/main/java/net/opentsdb/servlet/resources/QueryRpc.java
manolama: yup, will clean it up better later on though.

PR #1203, opened 2018-05-11
COMMON:

manolama:
- Add the ProcessorFactory and move iterator components from the
  QueryNodeFactory so node factories can be a bit more generic.
- Add types() to the QueryIteratorFactory.
- Register a common config in the UT configuration.
- Create the MockTSDB class to make UTs a bit easier.

CORE:
- Tweak the classes to use the ProcessorFactory and new mock TSDB.


manolama:
Merged in b3dff2c95014606e7d93a31d733fefb540379860

PR #1208, opened 2018-05-17
COMMON:

manolama:
- Fix a typo in the UnitTestConfiguration class and have it call back any
  config watchers when the override is used.

CORE:
- Remove the default expiration time in the TSDBV2Pipeline. We should let the
  key gen calculate it.
- Tweak the CachingQueryExecutor to pull the instantiated key gen from the
  plugin cache. Save cycles!
- Tweak DefaultTimeSeriesCacheKeyGenerator to use the new Configuration class
  and use a callback to update it's expiration settings online.
- Modify DefaultTimeSeriesCacheKeyGenerator to look for the metric downsampler
  from split-metric queries.


manolama:
Merged in e9e6c4ca97b7918e02675763edbf130aebbd4e8d.

PR #1212, opened 2018-05-19
COMMON:

manolama:
- Add a config key to enable auth in the Authentication class.

CORE:
- Register the auth key in the DefaultTSDB.

SERVLET:
- Add an AuthFilter interface to implement Authentication for servlets using
  the standard filter.
- Pull out the AuthState in the QueryRPC class. TODO - find a better way.

UNDERTOW:
- Add straight PEM parsing for certificates to avoid the debacle that is the
  Java KeyStore.


manolama:
Merged in a1077ddb5756f4045df7a0ec49eef666c0830f02

PR #1217, opened 2018-05-23
CORE:

manolama:
- Change the PluginConfig to a builder so it's cleaner.
- Add default types and default implementations to PluginsConfig so that the
  base config is much cleaner for users.
- Load those default types and implementations on initialization in the
  PluginsConfig.
- Have JsonV2QuerySerdes implement the TSD plugin interface.


manolama:
Merged in 8062ed3ee4d506f17ecc262f4eae384776a3e6d6, thanks!

PR #1221, opened 2018-05-26
COMMON:

manolama:
- Change the ASCII_CHARSET to ISO_8859_CHARSET which is what it really is
  and add a US ASCII set.

CORE:
- Don't throw an exception on duplicate plugins being loaded, just shutdown the
  dupe.
- Add a temp empty ctor to the JsonV2QuerySerdes to suppress plugin messages.
- Add the old and timeless salting configs to the Schema as well as load the
  default meta plugin.
- Add the row key salting calculation to the schema class. And modify the bucket
  salting method to avoid the byte instantiation.

ELASTICSEARCH:
- Exclude signatures from the shaded jar.

UNDERTOW:
- Log the auth filter loading.

STORAGE:
- Stub registering the multi-get properties in the HBase data store.
- Fix a bug in Tsdb1xMultiGet where it was incrementing the timestamp on the
  initial fetch so we'd miss the first hour of data. Also make sure we return
  the current result from a fetchNext() call.

### DIFF
--- a/core/src/main/java/net/opentsdb/storage/schemas/tsdb1x/Schema.java
+++ b/core/src/main/java/net/opentsdb/storage/schemas/tsdb1x/Schema.java
@@ -18,9 +18,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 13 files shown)


core/src/main/java/net/opentsdb/storage/schemas/tsdb1x/Schema.java
```diff
@@ -635,22 +646,72 @@ public int tagvWidth() {
...
     if (salt_width == 1) {
       row_key[0] = (byte) bucket;
       return;
     }
     
-    final byte[] salt = new byte[salt_width];
     int shift = 0;
     for (int i = 0; i <= salt_width; i++) {
-      salt[salt_width - i] = (byte) (bucket >>> shift);
+      row_key[salt_width - i] = 0;
```
mailravi3390: This will replace the first byte of the metric name when i = 0.
Also, total iterations are salt_width + 1. Starting from i = 1 may fix it.

manolama:
Merged in 2b020af6e504a4985e59475711027e1dbd22e6b7

PR #1227, opened 2018-06-05
CORE:

manolama:
- Handle a null trace tag in QueryRpc.

STORAGE:
- Return an empty result from the meta query node if the response was empty
  without fallback.


manolama:
Merged in b3160ff518911812680770ceff071df09a99e5e3

PR #1228, opened 2018-06-05
CORE:

manolama:
- Add a span to the MetaDataStorageSchema interface when running a query.
  Tracing is good.

SERVLET:
- Log exceptions passed through the ExceptionMappers.

STORAGE:
- Handle an edge condition in the mutli-get class if the results have already
  been passed up and the results are null.


manolama:
Merged in aa2631a6a3d61603db24ce00655ca0b5cc42f5d0

PR #1234, opened 2018-06-18
COMMON:

manolama:
- Add id() and type() methods to the QueryInterpolatorConfig interface.
- Define a register() method in the QueryInterpolatorFactory interface.
- Pass the QueryResult in the QueryIteratorFactory methods.
- Add a Span to the QueryNode initialize method.
- Make the QueryNodeConfig comparable and add a buildHashCode() method.
- Add a newNode() method to the QueryNodeFactory that allows for a default
  configuration and modify the other method to require a string ID for the
  node. Also add a nodeConfigClass() that we'll use for serdes on the configs.
- Require that QueryPipelineContext's return a TSDB instance and add a span
  to the initialize method.
- Add a RollupConfig method to the QueryResult class. Still need to figure out
  a better way to do this but for now it's the best spot for the config.
- Add QueryResults to the various source and processor factories.
- Rename the RollupConfig in core to DefaultRollupConfig and start an interface
  in common with the basics we need.
- TimeSeriesDataSource no longer extends QueryNodeFactory. It's a different
  thing that we have to support in a different factory when building an
  execution graph.
- Have TimeSeriesQuery return a HashCode from Guava and it must be comparable.

CORE:
- Disable the defaults in DefaultRegistry that would load the old junk we need
  to clean out.
- Load all QueryNodeFactories we can find, so it'll load the processors.
- Tweak the numeric fill policies to take a QueryInterpolatorConfig now and
  cast to the proper type.
- AbstractQueryNode now stores the ID and runs a span on initialization.
  Also implements equals() based on the ID and class.
- AbstractQueryPipelineContext is re-worked to compile from an ExecutionGraph
  that defines how the pipeline runs. It now loads node factories from the
  registry and creates instances of each node based on the factory.
- Add a BaseQueryNodeConfig class and a base class with interpolators to be
  used by the downsample and group by processors.
- Add the QueryDataSourceFactory. Execution graphs can specify a type of
  "DataSource" that will load this then use the node ID to pick the store to
  pull data from.
- Merge TSDBV2Pipeline into DefaultQueryContextBuilder and rename it
  TSDBV2QueryContextBuilder.
- Rework the ExecutionGraph and ExecutionGraphNode to be simple POJOs that
  we pass to a pipeline for construction. This is the start of merging the
  old execution code with the newer QueryNode code.
- Add a BaseInterpolatorConfig class and BaseQueryInterpolatorFactory along
  with the DefaultInterpolatorFactory that are now the registered factories
  used to get interpolators for data types. This fixes the v2 query pipeline
  where new factories were instantiated on every query. And now new data types
  can register interpolators for their types.
- Fix the BaseQueryNodeFactory to take the result and use the proper node name
  when registering a factory.
- Rework processors to pull the interpolators from factories instead of the
  config. Also, if a numeric summary config isn't present we'll use the numeric
  config.

SERVLET:
- Don't log exceptions on 404s.
- Fix a query bug by joining on the serdes step. TODO - find something better.

STORAGE:
- Return the base HBase data store as a parent instead of a factory since it
  isn't any more.
- The storage Tsdb1xQueryNode is now a full node, can't extend
  AbstractQueryNode since the parent isn't a factory any more.
- Check for empty rollup intervals in the scanners class.


manolama:
Merged in b742090b1b1ba1f93681f1c9604d349185be8eb2

PR #1238, opened 2018-06-23
COMMON:

manolama:
- Move the query interpolation interfaces to their own package. Also add
  a Jackson resolver so we can specify the type in JSON or YAML by setting
  the configType() value.
- Add config override methods to the QueryNodeConfig class. This may have to
  go but for now is useful.
- Add upstreamOfType() and downstreamOfType() to the QueryPipelineContext so
  nodes can find out how it may be able to optimize data for upstream.
- Add the QuerySinkConfig and QuerySinkFactory interfaces.
- Add ID and type to the SerdesOptions interface and stub a SerdesFactory
  interface.

CORE:
- Load the new SerdesFactory types by default in the registry.
- Implement the up and downstream types of methods in the
  AbstractQueryPipelineContext.
- Add the query overrides in the base query node configs.
- Modify QuerySourceConfig so that now it has the metric and timestamps.
  I'll probably move this around again later but now it's useful to decouple
  from a query type.
- Implement the SemanticQuery. This is a hopefully going to be the main query
  object in TSDB 3 that allows for a complete DAG to be defined and builds the
  execution graph. Also implement the SemanticQueryContext.
- ExecutionGraphNodes can store the configs directly so now the
  TSDBV2QueryContextBuilder will use those.
- JSonV2QuerySerdes now uses start and stop options set in it's options class.
- Add the JsonV2QuerySerdesFactory class.
- Add the config overrides to the RateOptions POJO since it has to have them for
  now and can't override a base abstract.
- The Downsampler now walks the nodes to find the time ranges. TODO - need a
  better way.
- Move some functions from the DownsampleConfig into the DownsampeResult since
  the config should just be a static POJO. And add intervalAsString() to the
  config.
- The GroupBy now resolves tags to byte arrays asynchronously in the onNext()
  method so we don't bind up on a join().

SERVLET:
- Stub out a RawQueryRpc class that takes in the SemanticQuery and allows for
  maximum flexibility in querying with OpenTSDB. It's ugly though and needs
  some DSLs to wrap it up.

STORAGE:
- Shift the rollup aggregation into the node instead of in the multi-get and
  the scanners. Saves code. And now it looks for downsamplers upstream.


manolama:
Merged in 9bce6cc83a5c7793db8d1b81731769c04a60c973

PR #1245, opened 2018-07-03
COMMON:

manolama:
- Add a flag to TimeSeriesByteId to indicate whether or not we should
  decode the metric or if it's been overwritten by an expression.

CORE:
- Implement the skip flag in the Base byte ID and handle it in the
  Schema class.
- Add a UT for Schema ID resolution.


manolama:
Merged in 7b1b2515e7f252f1ab8b4cedf0f2b95f78533c3e

PR #1246, opened 2018-07-03
COMMON:

manolama:
- Split the QueryNodeFactory into Single and Multi as we'll need the multi
  for expressions and other types of nodes.

CORE:
- Tweak AbstractQueryPipelineContext to handle the split node factories to
  allow nodes to be inserted in a graph.


manolama:
Merged in 0d0acae0022b20e1668cb11c994f80b46c604b53

PR #1249, opened 2018-07-06
COMMON:

manolama:
- Add a Comparators class to use with the Guava Comparison Chain.

CORE:
- Re-work the ExpressionConfig class (and rename it).
- Add hash/equals/compare to the Interpolator config classes.
- Remove the old Jexl binder classes and Jexl import now that we're getting
  our own grammar.


manolama:
Merged in fbece68b37329299ebe19942d8b0f4c8782daef8

PR #1252, opened 2018-07-09
CORE:

manolama:
- Modify the Joiner to simply prefer the alias instead of requiring it. Need that
  for sub-expressions.


manolama:
Merged in 74cb4506cab774e78b926abe44bcb8db1d5db705

PR #1255, opened 2018-07-11
CORE:

manolama:
- Add ExpressionNumericSummaryIterator to handle summary data. NOTE it's not
  great yet as it simply performs the same operation on all summaries in the
  data flowing through.


manolama:
Merged in be67152f45e473105eae71e0890c2838c5e04390

PR #1256, opened 2018-07-11
COMMON:

manolama:
- Add an overrideSource() method to the ExecutionGraphNode so that we can later
  on, dedupe configs for complex expressions in a single query. Not using it yet.

CORE:
- Add an equals method to the NumericLiteral in ExpressionParser.
- Add a BaseMultiQueryNodeFactory for processors that generate multiple nodes.
- Add the ExpressionFactory to populate a graph with expression nodes.


manolama:
Merged in 3bba9f6d07a4d5fea1f23852bd34349525a2ac76

PR #1258, opened 2018-07-12
CORE:

manolama:
- Remove the requirement that the query is st in the builder for
  QuerySourceConfig as deserialization of semantic queries won't have it.
- Fix Joiner bug with NATURAL and null encoded tags.
- Fix the BinaryExpression node to NOT complete upstream on responses and
  to only send the raw metric names for resolution in byte IDs.
- Fix the ExpressionTimeSeries to cast the factory properly and handle null
  time series. (and add UTs)


manolama:
Merged in 633a72f7cf562e1e9ffbac2f12b30b778ddf304d

PR #1259, opened 2018-07-12
COMMON:

manolama:
- Add bracket in the ExecutionGraphNode toString()

CORE:
- Tweak the Join code to use the "as" name for the metrics regardless of
  join type. Much cleaner for matching on metrics in expressions.
- Add "as" as the new name for a metric in the ExpressionConfig. Defaults to
  the expression ID.
- Fix the ExpressionFactory to walk the node graph and find and validate the
  expression variable names. It will replace the names when appropriate to
  make sure we match the proper metrics.


manolama:
Merged in 058100a2377d3e1fc9bb9244b8288768b4678103

PR #1262, opened 2018-07-14
CORE:

manolama:
- Implement the TopN processor by adding a config class with an aggregation function,
  count and top or bottom flag. Also add the numeric and numeric summary aggregators.


manolama:
Merged in fed9df5ad4cb59cbf0ab3013850cf04531a26def

PR #1263, opened 2018-07-16
COMMON:

manolama:
- Add TimeSeriesDatum interfaces for IDs and values. These will be used for
  writes as we only need namespace, metric and tags. Not the aggregated or
  disjoint tags.
- Add TimeSeriesDatumStringWrapperId that wraps a datum Id and returns a
  time series ID. This is useful for data stores.
- Add the TimeSeriesDatumIterable class, representing an iterable set of
  datum for writes as in a batched write request.
- Add the WritableTimeSeriesDataStore interface and rename the
  TimeSeriesDataStore to ReadableTimeSeriesDataStore. Some stores me be
  read-only for TSDB and others write-only so it make sense to break this
  up.

CORE:
- Add BaseTimeSeriesDatumStringId as a base class for the datum IDs.


manolama:
Merged in 82163692154f06d85a4c6cd8d5c0bd3ffe1131b5

PR #1265, opened 2018-07-17
COMMON:

manolama:
- Add the DatumIdValidator interface.

CORE:
- Add the DefaultDatumIdValidator validator that takes the place of the older
  Tags.java methods. It now properly checks to see if a string contains ASCII
  characters only (previously you could sneak in Unicode) and is much more
  flexible with it's configuration allowing for UTF8, whitespace and special
  ASCII chars.


manolama:
merged in a4498c6028f156348177cfb315bace1def3c5164

PR #1266, opened 2018-07-17
COMMON:

manolama:
- Move WriteState to a WriteStatus interface and implement a type for each.
  This will contain the error message users need for debugging writes.
- Change the WritableTimeSeriesDataStore to use the new WriteStatus.


manolama:
Merged in 291196d64d0e4067459e5c4e04a7a085e0e946ed

PR #1267, opened 2018-07-18
CORE:

manolama:
- Add the WriteState to IdOrError so we can figure out how to handle UID assignment
  failures.
- Restore the UID assignment to Tsdb1xUniqueIdStore using the old UniqueIdAllocator.
  However it now returns the IdOrError and also incorporate retry backoff using the
  same schema as AsyncHBase.


manolama:
Merged in d5e99a0a38b01665c8efa1e7556814b9530f0882

PR #1297, opened 2018-08-04
CORE:

manolama:
- Fix parsing of the 2x style downsampling where we were always returning null fills.


core/src/main/java/net/opentsdb/query/pojo/TimeSeriesQuery.java
```diff
@@ -762,7 +762,7 @@ private ExecutionGraphNode downsampler(final Metric metric,
     DownsampleConfig.Builder ds = (DownsampleConfig.Builder) DownsampleConfig.newBuilder()
         .setAggregator(downsampler.getAggregator())
         .setInterval(downsampler.getInterval())
-        .setFill(downsampler.getFillPolicy() != null ? true : false)
+        .setFill(policy != null && policy != FillPolicy.NONE ? true : false)
```
smrutilal2: `policy != null && policy != FillPolicy.NONE` is enough. No need of `?true:false`

PR #1302, opened 2018-08-07
COMMON:

manolama:
- Add WritableTimeSeriesDataStoreFactory.

CORE:
- Register the write and read stores in the default registry again. This may
  still need more tweaking as it's ugly as heck.
- Rework the order of default registrations in the plugin config as we need
  to load the built-ins first before named classes.
- Schema now implements the WritableTimeSeriesDataStore and registers the
  read writes.

PROTOBUF:
- Implement millisecond encoding for timestamps and fix zoned encoding.

BIGTABLE:
- Clean out some old printlns.
- Set the thread pool to be a fixed size. Cached can blow up the heap
  due to all of the async calls.

PUBSUB:
- Get the default write store for now.


manolama:
Merged in e259285d557033e342dd93e7426ba4ce1060d769

PR #1303, opened 2018-08-07
COMMON:

manolama:
- Add stringInterval() to the TimeSpecification for downsampling so we get
  the simple string, e.g. "1m" instead of the ISO "PT1M".

CORE:
- Fix Downsample to return the time spec ONLY if fill was enabled.


manolama:
Merged in a27f4d2d82aeececa2c70f897b47fe53eed560d2

PR #1311, opened 2018-08-10
CORE:

manolama:
- Add a flag in the QuerySourceConfig to fetch the last value from storage in
  the time range.


manolama:
Merged in 86c1965227ca65e8fd1fdc0b1447380e151cc899

PR #1312, opened 2018-08-12
COMMON:

manolama:
- Add the subtract() method to the TimeStamp interface.

CORE:
- And a starting offset to the NumericAggregator interface and fix up all
  of the methods.
- Fix the BaseQueryNodeFactory to return a TypedIterator for the other
  interface.
- Add the SlidingWindow node so we can now run the moving average and other
  functions.


manolama:
Merged in 94e8492fc633463f8779676c4efd0ba06d7e75f1

PR #1324, opened 2018-08-29
CORE:

manolama:
- Add the NumericArrayType for returning arrays of numerics up the stack. This
  can be much faster for some operations like GroupBy.
- Add a simple implementation in NumericArrayTimeSeries.


manolama:
Merged in f75bf236dfd53bd136fc41aececb4a16175eb16d

PR #1334, opened 2018-09-08
CORE:

manolama:
- Start a new DefaultQueryPlanner class that will parse out the ExecutionGraph
  and handle push-down operations to the data source. This will replace the
  old planners and the abstract context planner.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/AbstractQueryPipelineContext.java
+++ b/core/src/main/java/net/opentsdb/query/AbstractQueryPipelineContext.java
@@ -18,20 +18,14 @@ import java.time.temporal.ChronoUnit;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 
-import org.jgrapht.experimental.dag.DirectedAcyclicGraph;
-import org.jgrapht.experimental.dag.DirectedAcyclicGraph.CycleFoundException;
 import org.jgrapht.graph.DefaultEdge;
 import org.jgrapht.traverse.BreadthFirstIterator;
-import org.jgrapht.traverse.DepthFirstIterator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.reflect.TypeToken;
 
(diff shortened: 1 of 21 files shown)


core/src/main/java/net/opentsdb/query/AbstractQueryPipelineContext.java
```diff
@@ -59,7 +52,7 @@
  * 
  * <b>Invariants:</b>
  * <ul>
- * <li>Each {@link ExecutionGraphNode} must have a unique ID within the graph.</li>
+ * <li>Each {@link ExecutionGraphNode} must have a unique ID within the plan.graph().</li>
```
SeanPMiller: Are these invariants checked statically (can't find the code yet), or do they cause run-time errors?

core/src/main/java/net/opentsdb/query/execution/serdes/JsonV2QuerySerdes.java
```diff
@@ -470,4 +333,152 @@ public String version() {
...
+    }
+    json.writeEndObject();
+    json.writeArrayFieldStart("aggregateTags");
+    for (final String tag : id.aggregatedTags()) {
+      json.writeString(tag);
+    }
+    json.writeEndArray();
+    json.writeObjectFieldStart("dps");
+    
+    if (iterator.getType() == NumericType.TYPE) {
```
SeanPMiller: Hmm... moved the branching on the numeric type to here. Looks good.

core/src/test/java/net/opentsdb/query/TestAbstractQueryPipelineContext.java
```diff
@@ -416,154 +422,146 @@ public void initializeNoSuchUpstreamNode() throws Exception {
...
-    
-    assertNull(n1.config());
-    assertNull(n2.config());
-    assertNull(n3.config());
-    assertSame(config_a, n4.config());
-    assertNull(n5.config());
-    assertNull(s1.config());
-    assertNull(s2.config());
-  }
+//  @Test
```
SeanPMiller: Reintroducing this later?

core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
```diff
@@ -484,4 +259,258 @@ public void deserialize(final SerdesOptions options,
         + "class: " + getClass().getCanonicalName()));
   }
 
+  private void serializeSeries(
```
SeanPMiller: Same changes as V2...

core/src/main/java/net/opentsdb/query/AbstractQueryPipelineContext.java
manolama: Yes, throws a runtime error in the new `DefaultQueryPlanner` when it tries to build the graph.

core/src/test/java/net/opentsdb/query/TestAbstractQueryPipelineContext.java
manolama: yeah, will move it to the planner UTs.

manolama:
Merged in 4e3604bcbc51cbfcc6bf6bfd315975cf2bdf4323

PR #1341, opened 2018-09-16
CORE:

manolama:
- Load the array aggregators by default.
- Pass a reference to the GroupByResult to it's series instead of the source
  array.


manolama:
Merged in 2025be71903fff9da787387cb4e54c558dc619e3

PR #1345, opened 2018-09-18
CORE:

manolama:
- Run the GroupByNumericArrayIterator aggregation in the itertator's constructor
  so that we can release the reference to downstream iterators and avoid keeping too
  many arrays in memory.


manolama:
Merged in e79b1bde061f7fa0230d60da896e21c221007eb7

PR #1346, opened 2018-09-18
CORE:

manolama:
- Accumulate results in the AbstractQueryPipelineContext and close them *after*
  wrapping up with the time series. This is for results that may populate a span
  with stats on close after iterting.


manolama:
Merged in 9fcd7570a1714f514e82e083e4ff5276de0b1211

PR #1361, opened 2018-10-03
COMMON:

manolama:
- Rename TypedIterator to TimeSeriesIterator and make it an interface.
- Change the TimeSeries iterator() method to return a TypedTimeSeriesIterator
  instead of the plain iterator. Now it's like the iterators() call.
- Pass the type to QueryIteratorFactory's newIterator() and ProcessorFactory.
- Add sorting as a flag to the MockTimeSeries class.

CORE, PROTOBUF
- Implement the name change.


manolama:
MErged in a6b0bf93bc22a7d4131a559bd997e2205f4037d8

PR #1364, opened 2018-10-04
SERVLET:

manolama:
- Fix up the error handling by adding static serializers to the ExceptionMappers
  and make sure to log the completing query in the sink.
- Also clean up the async listeners in the RPC classes and get rid of the old
  async dispatch.


manolama:
Merged in 7153170d14c15d4e2968c33f34bbf1ac1857a8fd

PR #1365, opened 2018-10-04
SERVLET:

manolama:
- More tracing for 2x query.

BRAVE:
- Log the start thread in each child span.


manolama:
Merged in 82c145c014f10d0da873d2384bfc48214f378c74

PR #1366, opened 2018-10-05
CORE:

manolama:
- Losen up some checks and unmodifiables in the BaseTimeSeriesStringId ctor.
- Add getters to the BaseTimeSeriesStringId.Builder so we can avoid using
  separate objects in the merger builder.
- Add a flag to MergedTimeSeriesId to determine if we should perform a full
  merge with disjoin tags or just do a simple merge with the aggregated tags
  as per 2.x. The full compute turned out to be really expensive.
- Add a null check in the QuerySourceConfig for the source query.
- Fix an init but in TSDBV2QueryContextBuilder when filters are not supplied.
- Add the filters to data sources in the TimeSeriesQuery converter.


manolama:
Merged in f19f9bff057bca017b8d7801d31f4409077145d2

PR #1375, opened 2018-10-11
CORE:

manolama:
- Remove the groupAll flag and just use an empty tag list as the group all
  setting. Simpler.


manolama:
Merged in a641ba063e8015ec324dc67ded3e5b2cf902ad8c

PR #1377, opened 2018-10-11
CORE:

manolama:
- Add a setting tsd.query.convert.2x.rate1to0 that will help fix backwards compatible
  queries for the fixed rate behavior.


manolama:
Merged in 2d96a8640bd1c9f6b5583ff14f7d802a94d41ecb

PR #1381, opened 2018-10-16
COMMON:

manolama:
- Add some map getters to the registry so we can display information.

SERVLET:
- Add RegistryRpc so we can print out info about the plugins that are loaded.

### DIFF
--- a/common/src/main/java/net/opentsdb/core/Registry.java
+++ b/common/src/main/java/net/opentsdb/core/Registry.java
@@ -14,6 +14,7 @@
 // limitations under the License.
 package net.opentsdb.core;
 
+import java.util.Map;
 import java.util.concurrent.ExecutorService;
 
 import com.google.common.reflect.TypeToken;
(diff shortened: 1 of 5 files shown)


common/src/main/java/net/opentsdb/core/Registry.java
```diff
@@ -74,6 +75,9 @@ public void registerPlugin(final Class<?> clazz,
    */
   public <T> T getPlugin(final Class<T> clazz, final String id);
 
+  /** @return An umodifiable map of registered plugins. */
+  public Map<Class<?>, Map<String, TSDBPlugin>> plugins();
```
smrutilal2:
```suggestion
  public Map<Class<? extends TSDBPlugin>, Map<String, TSDBPlugin>> plugins();
```

manolama: good point will have to tighten it up under the hood a bit later

manolama:
Merged in db863b149e8cd32209596107d656d00fe22cc4f1

PR #1382, opened 2018-10-18
ALL:

manolama:
- Fix up all the classes to use the new QueryNodeFactory interfaces.
  Includes cleaning up some methods to use the BaseTSDBPlugin.

COMMON:
- API change for TSDBPlugin wherein we now have an id and a type. The id is now
  set at initialization so that we can instantiate multiple instances of the
  same type but with different IDs and configs. Allows for different data
  stores. (and fix the missing header)
- Move BaseTSDBPlugin to common (and at a UT class).
- Yank the *store* methods from Registry. We'll have a factory for each
  instance that returns the proper write/read node.
- Add TimeSeriesDataSourceFactory and drop the old ReadableTimeSeriesDataStore
  and the old TimeSeriesDataStoreFactory. A bit cleaner now.
- Drop the id from the QueryNodeFactory newNode() methods. We weren't using it
  and the IDs should come from the factory now.
- Create TimeSeriesDataSourceConfig as an interface with a builder. And drop
  the nested query (as it can lead to ugly circular references).
- Fix a missing copyright header in the QueryPlanner interface and add a
  reference to the context. And add java docs.

COMMON:
- Clean out some cruft in the DefaultRegistry.
- in the PluginConfig, now register ProcessrFactory and
  TimeSeriesDataSourceFactory factories instead of QueryNodeFactory as these
  are different types and I don't want dupes of the source factories
  registered as node factories.
  Also register a plugin with it ID if it's set to non-null. This is how we
  can use multi-colo to get at different data sources.
- Move QuerySourceConfig to BaseTimeSeriesDataSourceConfig as an abstract base
  and add DefaultTimeSeriesDataSourceConfig as a default implementation we can
  use with HBase, Bigtable and MockDataStore.
- Delete the QueryDataSourceFactory Now each store has to provide a factory to
  parse it's configs.
- Add parsing of sourceId from the source config and load default store
  factories in SemanticQuery when parsing the JSON.
- DefaultQueryPlanner will now parse out the sourceId from source configs and use
  proper data source factory.
- Fix missing header on LRUUniqueIdFactory.
- Add MockTSDBDefault for a singleton test for UTs to reduce execution time
  spent loading all of the classes. Use it!!


manolama:
Merged in 494f2e874c55e3dd94231082aedd3ae161b8f2e5

PR #1383, opened 2018-10-18
UNDERTOW:

manolama:
- Add Stijn de Witt's cors handler to Undertow and use almost the old settings
  from TSDB 2.x

### DIFF
--- a/implementation/server-undertow/src/main/java/net/opentsdb/tsd/TSDMain.java
+++ b/implementation/server-undertow/src/main/java/net/opentsdb/tsd/TSDMain.java
@@ -61,7 +61,6 @@ import io.undertow.Handlers;
 import io.undertow.Undertow;
 import io.undertow.Undertow.Builder;
 import io.undertow.server.HttpHandler;
-import io.undertow.server.handlers.PathHandler;
 import io.undertow.server.handlers.encoding.EncodingHandler;
 import io.undertow.servlet.Servlets;
 import io.undertow.servlet.api.DeploymentInfo;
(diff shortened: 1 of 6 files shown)


implementation/server-undertow/src/main/java/net/opentsdb/tsd/TSDMain.java
```diff
@@ -229,15 +243,24 @@ public static void main(final String[] args) {
...
+      if (!Strings.isNullOrEmpty(tsdb.getConfig().getString(CORS_PATTERN_KEY))) {
+        handler = new com.stijndewitt.undertow.cors.Filter(handler);
+        ((com.stijndewitt.undertow.cors.Filter) handler).setUrlPattern(
+            tsdb.getConfig().getString(CORS_PATTERN_KEY));
+        ((com.stijndewitt.undertow.cors.Filter) handler).setAllowHeaders(
+            tsdb.getConfig().getString(CORS_HEADERS_KEY));
+      }
+      
+      // support compression!
+      handler = new EncodingHandler.Builder()
```
gskchaitanya: Might be a good idea to expose the Undertow IO/workertask threads as a config

manolama: yeah I'll get a ton of configs in a future pr.

manolama:
Merged in 0487d76dfeb10267a9137f033d30e65d6bc8a31b

implementation/server-undertow/src/main/java/net/opentsdb/tsd/TSDMain.java
gskchaitanya: Right on!

PR #1385, opened 2018-10-19
switch start and end times if start > end

zb-oath:


core/src/test/java/net/opentsdb/query/TestSemanticQuery.java
```diff
@@ -123,6 +124,137 @@ public void builder() throws Exception {
...
+        .addSerdesConfig(JsonV2QuerySerdesOptions.newBuilder()
+            .addFilter("ds")
+            .setId("serdes")
+            .build())
+        .build();
+
+    assertEquals(null, endTimeBeforeCurrentTimeAndNoStartTime.getStart());
+    assertEquals(1000000000, endTimeBeforeCurrentTimeAndNoStartTime.startTime().epoch());
+    assertEquals("1000000000", endTimeBeforeCurrentTimeAndNoStartTime.getEnd());
+    assertEquals(DateTime.currentTimeMillis() / 1000, endTimeBeforeCurrentTimeAndNoStartTime.endTime().epoch());
```
manolama: We'll need to mock these out or it can be indeterminate if the host is really really slow. As an example, core/src/test/java/net/opentsdb/query/execution/TestTimeBasedRoutingExecutor.java

zb-oath: Are there any working examples?

manolama:
Merged in 202c3ac6615a694b22fe0b58fc4b1eef8ee7ed79

PR #1388, opened 2018-10-22
COMMON:

manolama:
- Remove the JGrapht dependency as I ran into a bug with the cycle detection.
- Bump to Guava 23.6.1-JRE for the newer graph functions.

CORE:
- Replace all of the graph code with Guava's implementation.
- Fix the expressions configs so that they reflect the regular nodes
  behavior and fix the naming of metrics so sub expressions return
  their IDs for matching upstream.

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -20,17 +20,6 @@
 
     </properties>
     
-    <dependencyManagement>
-      <dependencies>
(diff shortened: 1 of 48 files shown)


manolama:
Merged in 881a67f554011a92a54fc53de10ea031af2779c5

PR #1395, opened 2018-10-31
CORE:

manolama:
- Add the ByteToStringIdConverter that we can now use to convert
  result sets cleanly. Will let us yank a lot of code when we put it
  in the query planner.


manolama:
Merged in 030c8ebcfa3e113e9fdbbb4169924eb6ccc7f7f9

PR #1399, opened 2018-11-01
SERVLET:

manolama:
- Temporary fix that allows us to serialize exceptions properly again for
  the RPCs. The old way would send a 204 if something went pear shaped
  which is essentially useless.


manolama:
Merged in 1cd8d97379f85354a31e1f5928162d3a53a55668

PR #1403, opened 2018-11-04
COMMON:

manolama:
- Add a Json and Yaml file provider that parses those types and supports
  dotted paths and complex objects. We'll use it for the plugins config.
- Rename PropertiesFileFactory to FileFactory and supports the Json/Yaml
  files.


manolama:
Merged in 0b139386103e3f52becb630a7636c708bf5b86b9

PR #1422, opened 2018-11-15
CORE:

manolama:
- Fix the JSON parsing for TimeRouterConfigEntry.
- Also fix that entry so that it looks up the source ID factory of the ENTRY
  instead of the config.

HTTP:
- Don't send the upstream source ID down. TODO - make it configurable by the
  factory.


manolama:
Merged in 1cf5363120505c21881aad6e369e6d2165b5155a

PR #1424, opened 2018-11-16
CORE:

manolama:
- Add the source ID to the equals and hash code for
  BaseTimeSeriesDataSourceConfig as we need that to differentiate the graph.
- Fix the HAClusterFactory so that it can receive a non-HAClusterConfig
  and function. This is the case with the router.


manolama:
Merged in a3b12545b3159a9e8028fed0501334b132a598bc

PR #1425, opened 2018-11-17
PARENT:

manolama:
- Bump to Junit 4.12

COMMON:
- Add a logging interface to the QueryContext so that we can troublshoot
  queries easier and let users know if something goes wrong.
- Add a LogLevel flag to the TimeSeriesQuery so each call can determine
  the data to log.

CORE:
- Move a bunch of code from SEmanticQueryContext into a base context implementation.
- Remove the old TSDBV2QueryContextBuilder since we don't need it.
- Serialize the logs in the v3 serializer.
- Add a dummy log entry in the mock store.


manolama:
Merged in 23c0ac925479bb6a866c9b69774fcb405edac0db

PR #1429, opened 2018-11-23
Add systemd unit template to the RPM build

noharm:
This is my first attempt to add a systemd unit to be included in the RPM package. As the unit file is a template to launch multiple instances, I also modified logback.xml and the linked init.d script. I could also add a static opentsdb systemd unit file if desired, but the same can be accomplished by enabling the template as "opentsdb@4242.service". Please let me know what you think and if / how this can be improved.


johann8384:
Yes, please rebase this against next Thank you!

noharm:
Created #1450 against "next" to replace this PR

PR #1438, opened 2018-11-30
SERVLET:

manolama:
- Fix the async timeouts so they return proper JSON errors and now we
  set the timeout properly for the asyn by starting the query initialization
  in the context's start method.
  Thanks to @gskchaitanya for the timeout fix.


manolama:
Merged in e99df8e5899c958b0c732950fcacc3322a3054db

PR #1450, opened 2018-12-04
Add systemd unit template to the RPM build

noharm:
Add a systemd unit to be included in the RPM package. As the unit file is a template to launch multiple instances, I also modified logback.xml and the linked init.d script. I could also add a static opentsdb systemd unit file if desired, but the same can be accomplished by enabling the template as "opentsdb@4242.service". To replace PR #1429 against master


johann8384:
Does this still work in RHEL6?

PR #1451, opened 2018-12-05
CORE:

manolama:
- If a source has multiple upstream nodes, pass them up via the thread pool.
  TODO - This is just a quick hack to increase query parallelism. We need
  something smarter.

ELASTICSEARCH:
- Log some more information and handle max cardinality for multigets.

SERVLET:
- More query logging.

ASYNCHBASE:
- More query logging.


manolama:
Merged in 8baee595149548987fdf478c08d3aaca4ddbb026

PR #1454, opened 2018-12-10
COMMON:

manolama:
- Add rollupConfig() to TimeSeriesDataSourceFactory so the planner can find
  out what rollups are supported.
- Add namespace() to the TimeSeriesDataSourceConfig as we'll be able to use
  that with systems that support it.
- Also add getRollupIntervals and getRollupAggregations so that the
  downsample config can plan out what rollups to get.
- And add pre and post padding to the data source config so we can tell the
  driver how much to get before and after the query.
- Have the QueryPlanner interface allow other callers to get the factory for
  a node and return a collection of source nodes for a node.
- Add the getIntervals() and getPossibleIntervals() interfaces to RollupConfig.

CORE:
- Implement the new bits in BaseTimeSeriesDataSourceConfig.
- Implement the new bits in DefaultQueryPlanner and clean it up a little,
  moving more calls into the getFactory() method.
- DownsampleFactory will now scan for sources and update them with padding and
  an optional rollup config.
- Fix some logging in the DefaultRollupConfig to trace and implement the new
  methods
- Load the rollup config in the SchemaFactory now and share it with the Schema.

ASYNCHBASE/BIGTABLE:
- Use the new rollup setup.

### DIFF
--- a/common/src/main/java/net/opentsdb/data/TimeSeriesDataSourceFactory.java
+++ b/common/src/main/java/net/opentsdb/data/TimeSeriesDataSourceFactory.java
@@ -24,6 +24,7 @@ import net.opentsdb.query.QueryNodeConfig;
 import net.opentsdb.query.QueryNodeFactory;
 import net.opentsdb.query.TimeSeriesDataSourceConfig;
 import net.opentsdb.query.TimeSeriesQuery;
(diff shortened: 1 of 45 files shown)


manolama:
Merged in 1b2d5bafa38d99f3f9964c93f18be79cdf299df9

PR #1456, opened 2018-12-12
Fix #1448 by improving preconditions in TSDB.addPointInternal()

### LINKED ISSUE #1448: timestamp overflow
We use 4 bytes to store timestamp in rowKey, and designed 13 digits for millisecond resolution and 10 digits for second resolution at present. However, our code can not distinguish millisecond from second resolution in some corner cases.

The max timestamp(second) we can store is ``4294967295``(10 digits). But the max number of 10 digits is ``9999999999``. If we submit a ``4294967296`` timestamp, tsdb will treat it as millisecond. 

```
  private Deferred<Object> addPointInternal(final String metric,
                                            final long timestamp,
                                            final byte[] value,
                                            final Map<String, String> tags,
                                            final short flags) {

    ...
    // if timestamp=4294967296, we will treat it as millisecond.   
    if ((timestamp & Const.SECOND_MASK) != 0) {
      // drop the ms timestamp to seconds to calculate the base timestamp
      base_time = ((timestamp / 1000) - 
          ((timestamp / 1000) % Const.MAX_TIMESPAN));
    } else {
      base_time = (timestamp - (timestamp % Const.MAX_TIMESPAN));
    }

    ...
 }
    
```

Although it is impossible to have this ``4294967296`` timestamp in this century, we still need to defend this in our code. 

Just improve following preconditions:
```
  private Deferred<Object> addPointInternal(final String metric,
                                            final long timestamp,
                                            final byte[] value,
                                            final Map<String, String> tags,
                                            final short flags) {
    // we only accept positive unix epoch timestamps in seconds or milliseconds
    if (timestamp < 0 || ((timestamp & Const.SECOND_MASK) != 0 && 
        timestamp > 9999999999999L)) {
      throw new IllegalArgumentException((timestamp < 0 ? "negative " : "bad")
          + " timestamp=" + timestamp
          + " when trying to add value=" + Arrays.toString(value) + '/' + flags
          + " to metric=" + metric + ", tags=" + tags);
    }
```
 
I'd like to create a pull request to master if it's necessary.

ZephyrGuo:
Please see  #1448 for more details.

### DIFF
--- a/src/utils/DateTime.java
+++ b/src/utils/DateTime.java
@@ -18,6 +18,7 @@ import java.util.Calendar;
 import java.util.HashMap;
 import java.util.TimeZone;
 
+import net.opentsdb.core.Const;
 import net.opentsdb.core.Tags;
 
 /**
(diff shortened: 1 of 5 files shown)


src/utils/DateTime.java
```diff
@@ -578,4 +579,17 @@ public static int unitsToCalendarType(final String units) {
     throw new IllegalArgumentException("Unrecognized unit type: " + units);
   }
 
+  public static boolean isValidTimestamp(final long timestamp) {
```
manolama: Javadocs to describe what the new methods do please.

ZephyrGuo: Done.

PR #1458, opened 2018-12-14
For branch next, add an expression function named FirstDifference, wh…

Eduardo95:
…ich calculates the first difference of a time series. I noticed there is MovingAverage calculation, so I thought maybe I can enrich some mathematics functions into that.


Eduardo95:
@manolama Unit test added, please check.

PR #1466, opened 2018-12-18
COMMON:

manolama:
- Add a getSecretObject() method to the SecretProvider and let the base
  implement a null getSetting() return.


manolama:
Merged in cb8dc8761a92d8e8c2eab60c05f56adca7971a68.

PR #1470, opened 2018-12-19
UNDERTOW:

manolama:
- Add an AuthFilter InstanceFactory so that the filter, instantiated already
  as a TSDB plugin, can be passed as-is to Undertow. Otherwise we'd have
  multiple instances of it.


manolama:
Merged in 112ed3946d7798834988b89de3c79cf8965dfd46

PR #1472, opened 2018-12-26
Do PutDataPointRpc.GroupCB asynchronously

ZephyrGuo:
There is a bottleneck that the Netty worker threads are the bottleneck for responses. 

Recently, we are doing pressure test for tsdb. We put data into a single tsd (1 tsd, 1 regionserver, batch 50, threads 48, in sync mode), and we found it can not run stably.  See the following picture.
tps-tsdb-1
TSDB's memory grow up quickly, and then cause FGC. The root cause is too many `PutRequest` in AsyncHBase. These `PutRequests` can not be removed from `rpcs_inflight`, because there is only one Netty worker decodes responses from regionserver (each regionserver corresponds to a NioClientSocketChannel). 

When AsyncHBase finish a RPC, it will remove this RPC from rpcs_inflight.  After that, AsyncHBase will call GroupCB, and then GroupCB send HTTP response to the client.  All of this runs in just one Netty worker. We tried to do GroupCB asynchronously, and got some good results. TSDB can run stably with almost 650K TPS.

performacne-tsdb-2

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -78,6 +78,7 @@ tsdb_SRC := \
 	src/core/RequestBuilder.java	\
 	src/core/RowKey.java	\
 	src/core/RowSeq.java	\
+	src/core/RpcResponder.java	\
 	src/core/iRowSeq.java	\
 	src/core/SaltScanner.java	\
 	src/core/SeekableView.java	\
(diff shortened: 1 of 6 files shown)


manolama:
LGTM. We'll need to pass queries through this as well.

PR #1475, opened 2019-01-02
HTTP:

manolama:
- Fix a parsing bug in the v3 results wherein if the array switches from a
  double to a long when serializing, it wouldn't properly handle the long
  and store it in the double array.
- Cleanup the V3 source and set the type to override others like HAConfig.
- Also pass in the log level and parse out the logs from downstream.


manolama:
Merged in 81f1380faf623045ed45ad687753c93ae9035ca0

PR #1478, opened 2019-01-04
Always write tool output (fsck, metasync, etc.) to STDOUT by default

johnseekins:
This helps make problems like https://github.com/OpenTSDB/opentsdb/issues/1351...less...

Basically, all CLI tools should write to stdout by default so people can see them working.


johnseekins:
I would _love_ to. But I'm terrible at git. Is that better?

johnseekins:
Nope. That made it worse.

PR #1480, opened 2019-01-08
CORE:

manolama:
- Fix TopN wherein Numbers aren't directly comparable by switching the map
  to a double.


manolama:
Merged in bce76e38a990a97236529560066b977a4fa5a83e

PR #1484, opened 2019-01-11
UNDERTOW:

manolama:
- Add tsd.network.tls.protocols and tsd.network.tls.ciphers to better protect
  the server.
- Also fix a bug if not using a secret TLS cert and key.


manolama:
Merged in 9d70130ff652d01851bdf4de34e77927d5b9d2e7

PR #1485, opened 2019-01-11
CORE:

manolama:
- Fix a long to double array bug in the downsampler.


core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericIterator.java
```diff
@@ -371,11 +371,10 @@ private void add(final double value) {
...
       if (double_values == null) {
         double_values = new double[long_values.length];
+      } else if (long_values.length > double_values.length) {
+        double_values = new double[long_values.length];
       }
-      if (value_idx == 0) {
-        return;
-      }
-      for (int i = 0; i < value_idx; i++) {
+      for (int i = 0; i < long_values.length - 1; i++) {
```
sergeykhegay: Why the last value is not copied?

sergeykhegay: I see that `shiftToDouble` is tightly coupled with `add`, both have to be called together in order to make sense for the off-by-one copying. Does it make sense to copy everything, or better up_to `value_idx` elements?

PR #1494, opened 2019-01-18
COMMON:

manolama:
- Add RefreshingSSLContext to build SSLContexts. TODO - needs tests and more
  validation.

HTTP-EXECUTOR:
- Add an opentsdb.http.executor.headers.user setting so that if a TSD is used
  as a proxy to other TSDs (e.g. HA Cluster) and we're using something like
  client certificates for authentication, we can send the real user downstream.
  TODO - more work around this.
- Rename SharedHttpClient to DefaultSharedHttpClient as the base implementation
  and create SharedHttpClient as an interface.
- In the v3 source, return the underlying error instead of the "unavailable
  hosts" error.

UNDERTOW:
- Use the new SSLContext refresher.

### DIFF
--- a/common/src/main/java/net/opentsdb/utils/RefreshingSSLContext.java
+++ b/common/src/main/java/net/opentsdb/utils/RefreshingSSLContext.java
@@ -1,0 +1,800 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2019  The OpenTSDB Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package net.opentsdb.utils;
+
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.StringReader;
+import java.math.BigInteger;
+import java.security.GeneralSecurityException;
+import java.security.KeyFactory;
+import java.security.KeyManagementException;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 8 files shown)


common/src/main/java/net/opentsdb/utils/RefreshingSSLContext.java
```diff
@@ -0,0 +1,800 @@
...
+import sun.security.util.DerValue;
+
+/**
+ * First stab at a refreshing SSL context that watches the given params for
+ * updates and reloads the context on changes. Note that if the interval is set
+ * to 0 then refreshing is disabled. Makes it useful for servers that load once.
+ * 
+ * @since 3.0
+ */
+public class RefreshingSSLContext implements TimerTask {
```
smrutilal2: Does the Cert refresher from Athenz not fit the bill?

manolama: It does, this is just in case we don't have Athenz and eventually I'll make it work like Athenz. I already wrote this code so may as well keep it.

manolama: Oh and eventually this will be useful for the Undertow context so we can renew the TLS cert without restarting.

smrutilal2: Athenz Cert refresher would work for Undertow as well. Too much of code. I would recommend not bother maintaining it.

manolama: No I mean the Athenz refresher is only good for client certs right? Not server certs?

executors/http/src/main/java/net/opentsdb/utils/DefaultSharedHttpClient.java
```diff
@@ -0,0 +1,286 @@
...
+
+/**
+ * A shared asynchronous HTTP client for making remote calls to various
+ * services.
+ * 
+ * @since 3.0
+ * @author clarsen
+ *
+ */
+public class DefaultSharedHttpClient extends BaseTSDBPlugin implements RefreshCallback {
```
smrutilal2: What do you mean by Shared here?

smrutilal2: Looks like just a name. Nothing special about Sharing here.

manolama: As it multiple sources (e.g. multi-colo) can use the same client pool instead of instantiating one per node.

manolama:
Merged in 3508ac7d3508f4bfe082277da399f51936b3ec0e

PR #1497, opened 2019-01-20
PARENT:

manolama:
- Add the git-commit-id-plugin to pull commit and version information at
  build time.

COMMON, CORE:
- Add version load and parse classes.

UNDERTOW:
- Report the version.

### DIFF
--- a/common/src/main/java/net/opentsdb/version/CommonVersion.java
+++ b/common/src/main/java/net/opentsdb/version/CommonVersion.java
@@ -1,0 +1,122 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2019  The OpenTSDB Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


common/src/main/java/net/opentsdb/version/CommonVersion.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.InputStream;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * A utility that pulls Version and Git information from the build using the
```
sidhhu: aren't there any libraries to do this already?

manolama: That's the Maven plugin I'm using :) Then I just have to pull out the info.

PR #1509, opened 2019-01-29
Explicit tags fix for 1.x scanners

gskchaitanya:


core/src/test/java/net/opentsdb/query/filter/TestFilterUtils.java
```diff
@@ -130,51 +130,51 @@ public void matchesTags() throws Exception {
...
+    assertTrue(FilterUtils.matchesTags(filter, tags, null));
     
     try {
-      FilterUtils.matchesTags(null, tags);
+      FilterUtils.matchesTags(null, tags, null);
       fail("Expected IllegalArgumentException");
     } catch (IllegalArgumentException e) { }
     
     try {
-      FilterUtils.matchesTags(filter, null);
```
manolama: Add a couple of tests here please :)

PR #1515, opened 2019-01-31
Update opentsdb_restart.py

ddoshy:
Use 'systemctl' rather than 'service'

### DIFF
--- a/tools/opentsdb_restart.py
+++ b/tools/opentsdb_restart.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 """Restart opentsdb. Called using -XX:OnOutOfMemoryError=<this script>
 
-Because it's calling the 'service opentsdb' command, should be run as root.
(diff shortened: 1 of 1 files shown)


tools/opentsdb_restart.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 """Restart opentsdb. Called using -XX:OnOutOfMemoryError=<this script>
 
-Because it's calling the 'service opentsdb' command, should be run as root.
+Because it's calling the 'systemctl *action* opentsdb' command, should be run as root.
```
manolama: Could you add a check to determine if an OS is running the old school *init* vs *systemd* please? We do have users on older systems.

manolama (dismissed):

tools/opentsdb_restart.py
johann8384: They would be 9 or 10 year old OS versions now so I'm just going to merge this.

PR #1525, opened 2019-02-06
Normalize tag support

gnydick:
This new plugin type allows you to provide a plugin to normalize tags. My use case for implementing this is the fact that the slew of Cloud Native systems coming online all export their metrics slightly differently and opentsdb doesn't support fuzzy matches or specifying which tags are equivalent to each other.

I was receiving container, container-name, container_name, etc., etc., etc.

This plugin type requires a normalizeTags(tags) method. The tags are passed in and a normalized set are returned.


configure.ac
```diff
@@ -14,7 +14,7 @@
 # along with this library.  If not, see <http://www.gnu.org/licenses/>.
 
 # Semantic Versioning (see http://semver.org/).
-AC_INIT([opentsdb], [2.4.0], [opentsdb@googlegroups.com])
+AC_INIT([opentsdb], [2.4.0-tnp], [opentsdb@googlegroups.com])
```
johann8384: We probably don't need this change.

src/core/TSDB.java
```diff
@@ -12,59 +12,17 @@
 // see <http://www.gnu.org/licenses/>.
 package net.opentsdb.core;
 
-import java.io.File;
```
johann8384: So, it looks like you did the "auto format" in IntelliJ, I've been tempted to commit this myself. Would it be possible to isolate just the changes relevant to your new plugin type?

I love the concept and would be happy to get this merged if we can remove the unrelated changes.

src/tsd/RTPublisher.java
```diff
@@ -28,7 +28,7 @@
  * after they are queued for storage. In the future we may support publishing
  * meta data or other types of information as changes are made.
  * <p>
- * <b>Note:</b> Implementations must have a parameterless constructor. The
```
johann8384: You can leave in all the "trailing whitespace" fixes if you want! :)

johann8384 (changes requested):

configure.ac
gnydick: Sounds good!

johann8384:
Merging in #2302. @gnydick I know it's been, um, years, but if you have an example of a working tag normalization plugin that would be great!

sins921 (approved):

gnydick:
Let me see if I have an example for you...

gnydick:
Here's an example plugin.  https://github.com/OpenTSDB/opentsdb/commit/6aba0eda2dcf3dea1f390b2484cd00915e649abc

sins921 (approved):

PR #1535, opened 2019-02-12
Honor rollups for auto interval and process Nan rollup values

gskchaitanya:


core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
```diff
@@ -511,17 +525,19 @@ private void writeNumericSummary(
...
             result);
         return;
       }
 
+      Collection<Integer> integers =
+          ((TimeSeriesValue<NumericSummaryType>) value)
+              .value()
+              .summariesAvailable();
+
+      final List<Integer> summaries = Lists.newArrayList(integers);
```
manolama: Why the dupe here? This may be super expensive as it would create a new list for every data point

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleFactory.java
```diff
@@ -219,11 +219,13 @@ public void setupGraph(final QueryPipelineContext context,
         .setStart(context.query().getStart())
         .setEnd(context.query().getEnd())
         .setId(config.getId());
-    
+
+    QueryNodeConfig newConfig = builder.build();
```
manolama: good catch.

PR #1548, opened 2019-02-20
CORE:

manolama:
- Fix another bug in the QueryPlanner where a concrete node could be skipped
  during initialization.
- Fix the Rate node to handle arrays.
  TODO - still not happy with the reset/counter logic. Something's still off
  here.

HTTP:
- Log success and emitting queries at Debug instead of Trace.

### DIFF
--- a/executors/http/src/main/java/net/opentsdb/query/execution/HttpQueryV3Source.java
+++ b/executors/http/src/main/java/net/opentsdb/query/execution/HttpQueryV3Source.java
@@ -326,8 +326,10 @@ public class HttpQueryV3Source extends AbstractQueryNode implements SourceNode {
                 .setDataSource(config.getId())
                 .build());
           } else {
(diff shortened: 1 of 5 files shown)


executors/http/src/main/java/net/opentsdb/query/execution/HttpQueryV3Source.java
```diff
@@ -326,8 +326,10 @@ public void completed(final HttpResponse response) {
                 .setDataSource(config.getId())
                 .build());
           } else {
-            LOG.trace("Successful reseponse from [" + host + endpoint + "] after " 
-                  + DateTime.msFromNanoDiff(DateTime.nanoTime(), start) + "ms");
+            if (LOG.isDebugEnabled()) {
+              LOG.debug("Successful reseponse from [" + host + endpoint + "] after "
```
sidhhu: typo `reseponse `

PR #1553, opened 2019-03-01
CORE:

manolama:
- Add a base pool allocator class.
- Add a dummy pool that simply returns new objects from the allocator.
- Add a long array and double array pool allocator.

STORMPOT:
- Remove the allocator reference and just use the config. Save a ref.


manolama:
Merged in b448ffada4ca007bf720c503bf9a05402713b25e

PR #1567, opened 2019-03-22
Add "check_tsd_v2" script

johnseekins:
This script performs somewhat similarly to the `check_tsd` script, but it evaluates each individual metric group separately when given a filter like `{host=*}`. This allows us to make slightly more fine-grained alerts.

It also allows for multiple expressions to be evaluated on the metric.


johann8384:
To be honest, I'm conflicted about merging this. One one hand I want to say "sweet!" and hit merge. On the other, I'm frustrated that we didn't just improve upon the existing check_tsd. Do you have any insight there?

johnseekins:
This returns data differently. I didn't want to break functionality of the original script if people were using it.

johnseekins:
Woo! Thanks, Jonathan!

PR #1582, opened 2019-04-03
fix #1581

### LINKED ISSUE #1581: Fix an edge case in TsdbQuery.getScanEndTimeSeconds()
```
    // Convert to seconds if we have a query in ms.
    if ((end & Const.SECOND_MASK) != 0L) {
      end /= 1000L;
      if (end - (end * 1000) < 1) {
        // handle an edge case where a user may request a ms time between
        // 0 and 1 seconds. Just bump it a second.
        end++;
      }
    }
```
The condition `end - (end * 1000) < 1` is always true in fact. So if we request `1554260399000`, we will get `1554267600`. The correct result is `1554264000`.

Expected end scan time: `1554260400 = 1554260399000/1000 - (1554260399000 / 1000 % 3600) + 3600`
Actual end scan time: `1554264000 = (1554260399000/1000 + 1) - ((1554260399000/1000 +1) % 3600) + 3600`

`1554264000` is one hour behind of `1554260400`.

We might scan one more row of data. If we delete data, which may lead to delete more.

ZephyrGuo:
Please see #1581 for more details.

### DIFF
--- a/test/core/TestTsdbQuery.java
+++ b/test/core/TestTsdbQuery.java
@@ -32,6 +32,8 @@ import net.opentsdb.storage.MockBase;
 import net.opentsdb.uid.NoSuchUniqueName;
 import net.opentsdb.utils.DateTime;
 
+import org.jboss.netty.util.internal.ThreadLocalRandom;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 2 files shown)


test/core/TestTsdbQuery.java
```diff
@@ -97,6 +99,24 @@ public void setEndTime() throws Exception {
     assertEquals(1356998400L, query.getEndTime());
   }
 
+  @Test
```
johann8384: Thank you for adding the test!

ZephyrGuo: Thank you for review.

PR #1593, opened 2019-04-10
EXECUTORS:  HA cluster mode, fix downsample interval

gskchaitanya:


executors/http/src/main/java/net/opentsdb/query/execution/HttpQueryV3Source.java
```diff
@@ -188,8 +189,20 @@ public void fetchNext(final Span span) {
...
+    for (QueryNodeConfig pushdown : config.getPushDownNodes()) {
+      if (pushdown instanceof DownsampleConfig) {
+        DownsampleConfig.Builder b = DownsampleConfig
+            .newBuilder((DownsampleConfig) pushdown)
+            .setStart(context.query().getStart())
+            .setEnd(context.query().getEnd())
+            .setId(pushdown.getId());
+
+        pushdown = b.build();
+        config.getPushDownNodes().set(index, pushdown);
```
manolama: Can't do this as it's a Concurrent Modification exception. Just push it directly to the new list as per line 204.

manolama: Guess it's ok since we're modifying the set position.

manolama:
I'll fix the UT in the next PR.

PR #1594, opened 2019-04-11
ELASTICSEARCH: Fix multicolo for meta when one colo returns a null re…

sidhhu:
…sponse


implementation/elasticsearch/src/main/java/net/opentsdb/meta/NamespacedAggregatedDocumentSchema.java
```diff
@@ -155,104 +155,100 @@ public String version() {
...
           MetaQuery meta_query = query.metaQueries().get(i);
           NamespacedAggregatedDocumentResult result = null;
-          for (final Map.Entry<String, MultiSearchResponse> search_response: results.entrySet()) {
-            final MultiSearchResponse.Item[] responses= search_response.getValue().getResponses();
+          int null_results = 0;
+          for (final Map.Entry<String, MultiSearchResponse> search_response : results.entrySet()) {
+            final MultiSearchResponse.Item[] responses = search_response.getValue().getResponses();
             final SearchResponse response = responses[i].getResponse();
             if (response == null) {
               LOG.warn("Null response from " + search_response.getKey());
```
arungupta: need to log which query had null response and have a metric for this

PR #1606, opened 2019-04-19
CORE: Fix downsample logic and add more downsample intervals

sidhhu:


core/src/main/java/net/opentsdb/query/SemanticQuery.java
```diff
@@ -347,6 +349,10 @@ public static Builder parse(final TSDB tsdb, final JsonNode root) {
         if (temp != null && !temp.isNull()) {
           config_factory = tsdb.getRegistry()
               .getQueryNodeFactory(temp.asText());
+          if (config_factory instanceof DownsampleFactory) {
```
manolama: Lets do this part in the downsample factory itself. It should have the full query from the QueryContext.

sidhhu:
will be fixed in #1593

PR #1614, opened 2019-04-24
CORE:

manolama:
- Add array count, first and last aggregators.

### DIFF
--- a/core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayFirstFactory.java
+++ b/core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayFirstFactory.java
@@ -1,0 +1,112 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2019  The OpenTSDB Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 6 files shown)


core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayFirstFactory.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public void accumulate(final long[] values, 
+                           final int from, 
+                           final int to) {
+      if (double_accumulator == null && long_accumulator == null) {
+        long_accumulator = Arrays.copyOfRange(values, from, to);
+        return;
+      }
+      
+      if (long_accumulator != null) {
```
sidhhu: why not have an inverted condition instead of having a no-op?
And, trying to understand this. you have an array of long values but why are you filling an array of doubles?

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayFirstFactory.java
```diff
@@ -0,0 +1,112 @@
...
+  
+  public static class ArrayFirst extends BaseArrayAggregator {
+
+    public ArrayFirst(final boolean infectious_nans) {
+      super(infectious_nans);
+    }
+
+    @Override
+    public void accumulate(final long[] values, 
+                           final int from,
```
sidhhu: do you need `from` and `to` for these aggregators? I'm assuming from would always be 0 and to as the length of array?

manolama: We need them for shared buffers in the future where a component gets just a slice of a big array.

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayFirstFactory.java
manolama: Yeah I do the inverted. The first array we get determines the final data array we return. If it's a long then we have a full array of values and we ignore everything else. If it was a double though, we need NaN handling to find the first "real" values.

PR #1617, opened 2019-05-01
CORE: Fix TimeShift queries

sidhhu:


core/src/test/java/net/opentsdb/query/processor/timeshift/TestTimeShiftNumericArrayIterator.java
```diff
@@ -0,0 +1,82 @@
+package net.opentsdb.query.processor.timeshift;
```
manolama: license header please.

PR #1618, opened 2019-05-01
COMMON:

manolama:
- Change PartialTimeSeries to return a TimeSeriesDataType.

CORE:
- Update Tsdb1xNumericPartialTimeSeries to use the new interface.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/hacluster/HAClusterFactory.java
+++ b/core/src/main/java/net/opentsdb/query/hacluster/HAClusterFactory.java
@@ -14,11 +14,6 @@
 // limitations under the License.
 package net.opentsdb.query.hacluster;
 
-import java.time.temporal.TemporalAmount;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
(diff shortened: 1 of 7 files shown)


core/src/main/java/net/opentsdb/pools/ByteArrayPool.java
```diff
@@ -0,0 +1,87 @@
...
+  public Deferred<Object> initialize(final TSDB tsdb, final String id) {
+    if (Strings.isNullOrEmpty(id)) {
+      this.id = TYPE;
+    } else {
+      this.id = id;
+    }
+    
+    registerConfigs(tsdb.getConfig(), TYPE);
+    length = tsdb.getConfig().getInt(configKey(LENGTH_KEY, TYPE));
+    size = (8 * length) +  + 16 /* 64-bit overhead */;
```
gskchaitanya: two + ?

manolama: arggg, good catch. and it's not a long so it's a single byte.

PR #1626, opened 2019-05-07
ELASTICSEARCH: use a composite key with id and namespace in map

sidhhu:


common/src/main/java/net/opentsdb/meta/NamespacedKey.java
```diff
@@ -0,0 +1,49 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2018  The OpenTSDB Authors.
```
manolama: 2019

common/src/main/java/net/opentsdb/meta/NamespacedKey.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @since 3.0
+ */
+public class NamespacedKey {
+
+  private String namespace;
+
+  private String id;
+
+  public NamespacedKey(String namespace, String id) {
```
manolama: finals/javadoc please

PR #1640, opened 2019-05-17
3x tweaks

manolama:


storage/asynchbase/src/main/java/net/opentsdb/storage/Tsdb1xMultiGet.java
```diff
@@ -87,56 +105,54 @@
...
+  protected List<byte[]> tsuids;
   
   /** The number of get requests batches allowed to be outstanding at
    * any given time. */
-  protected final int concurrency_multi_get;
+  protected int concurrency_multi_get;
   
   /** Whether or not we're fetching rows in reverse. */
-  protected final boolean reversed;
+  protected boolean reversed;
```
sidhhu: why?

manolama: Cause it's a pooled object now so this has to be reset on each query.

sidhhu: i mean why use reversed at all.

manolama: oh cause we will eventually stream data from most recent to oldest so we can load like Splunk does.

PR #1655, opened 2019-05-30
3x tweaks

manolama:


implementation/query-runner/src/main/java/net/opentsdb/tsd/QueryConfig.java
```diff
@@ -263,14 +277,14 @@ public void cancel() {
   class ResponseCallback implements FutureCallback<HttpResponse> {
     private final int index;
     private final TSDB tsdb;
-    private StatsTimer timer;
+    private volatile StatsTimer timer;
```
gskchaitanya: We need timer to be volatile? I see it being used in the synchronized methods only.

manolama: Yeah I can yank that now, thanks.

implementation/query-runner/src/main/java/net/opentsdb/tsd/QueryConfig.java
```diff
@@ -343,13 +359,169 @@ void stop(final int status_code) {
...
+        if (!Strings.isNullOrEmpty(runner.curl_flags)) {
+          // TODO - handle quotes if we need em.
+          final String[] parts = runner.curl_flags.split(" ");
+          for (int i = 0; i < parts.length; i++) {
+            args.add(parts[i]);
+          }
+        }
+        args.add(shuffled_endpoints.get(index));
+        builder.command(args);
+        builder.redirectErrorStream(true);
```
gskchaitanya: Nice, this would merge the error stream into inputstream so we don't have to consume that separately

PR #1665, opened 2019-06-04
Executors as a plugin

gskchaitanya:


common/src/main/java/net/opentsdb/threadpools/FixedThreadPoolExecutor.java
```diff
@@ -0,0 +1,132 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2018 The OpenTSDB Authors.
```
manolama: 2019

common/src/main/java/net/opentsdb/threadpools/FixedThreadPoolExecutor.java
```diff
@@ -0,0 +1,132 @@
...
+    if (!tsdb.getConfig().hasProperty(getConfigKey(QUEUE_MAX_SIZE))) {
+      tsdb.getConfig().register(getConfigKey(QUEUE_MAX_SIZE), 100000, false,
+          "The ID of an object pool factory plugin to use for this pool. "
+              + "Can be null to use the default.");
+    }
+
+    Integer maxSize = tsdb.getConfig().getInt(getConfigKey(QUEUE_MAX_SIZE));
+    LOG.info("Initializing new FixedThreadPoolExecutor with queue max capacity of {}", maxSize);
+
+    int nThreads = Runtime.getRuntime().availableProcessors() * 2;
```
manolama: Make this configurable too.

common/src/main/java/net/opentsdb/threadpools/FixedThreadPoolExecutor.java
```diff
@@ -0,0 +1,132 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Strings;
+import com.stumbleupon.async.Deferred;
+
+import net.opentsdb.core.TSDB;
+import net.opentsdb.query.QueryContext;
+
+public class FixedThreadPoolExecutor implements TSDBThreadPoolExecutor {
```
manolama: Java docs about how it's a super simple non-scheduling pool.

common/src/main/java/net/opentsdb/threadpools/UserAwareThreadPoolExecutor.java
```diff
@@ -0,0 +1,352 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2018 The OpenTSDB Authors.
```
manolama: 2019

common/src/main/java/net/opentsdb/threadpools/FixedThreadPoolExecutor.java
```diff
@@ -0,0 +1,132 @@
+// This file is part of OpenTSDB.
```
manolama: Oh and move implementations to the `core` package please.

common/src/main/java/net/opentsdb/threadpools/TSDBThreadPoolExecutor.java
```diff
@@ -0,0 +1,100 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2018  The OpenTSDB Authors.
```
manolama: 2019

common/src/main/java/net/opentsdb/threadpools/TSDBThreadPoolExecutor.java
```diff
@@ -0,0 +1,100 @@
...
+package net.opentsdb.threadpools;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
+
+import net.opentsdb.core.TSDBPlugin;
+import net.opentsdb.query.QueryContext;
+
+public interface TSDBThreadPoolExecutor extends TSDBPlugin {
```
manolama: java doc about what it's for.

common/src/main/java/net/opentsdb/threadpools/TSDBThreadPoolExecutor.java
```diff
@@ -0,0 +1,100 @@
...
+  public static final String KEY_PREFIX = "threadpool.";
+
+  public static final String QUEUE_MAX_SIZE = "queue.max.size";
+  public static final String CORE_THREAD_POOL_SIZE = "core.thread.pool.size";
+  public static final String MAX_THREAD_POOL_SIZE = "max.thread.pool.size";
+
+  public String tpType();
```
manolama: Java doc please.

common/src/main/java/net/opentsdb/threadpools/UserAwareThreadPoolExecutor.java
```diff
@@ -0,0 +1,352 @@
+// This file is part of OpenTSDB.
```
manolama: Move implementation to `core`.

common/src/main/java/net/opentsdb/threadpools/UserAwareThreadPoolExecutor.java
```diff
@@ -0,0 +1,352 @@
...
+    workQueue = new LinkedBlockingQueue<Runnable>(maxSize);
+
+    // TODO: better alternative?
+    // If the worker queue is full, we run the task in the thread that calls execute(task)
+    RejectedExecutionHandler inPlaceRunTaskHandler = new RejectedExecutionHandler() {
+
+      @Override
+      public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
+        if (!executor.isShutdown()) {
+          r.run();
```
manolama: Yeah thats ok I think but log if we skip the executor.
And throw an error if the executor is indeed shut down as that means we closed the executor too early.

common/src/main/java/net/opentsdb/threadpools/UserAwareThreadPoolExecutor.java
```diff
@@ -0,0 +1,352 @@
...
+
+    @Override
+    public void run() {
+
+      if (qctx == null) {
+        r.run();
+        return;
+      }
+
+      final String user = this.qctx.authState().getUser();
```
manolama: Check for nulls here. If they are null then use the string `Unkown` and we'll throttle on that.

common/src/main/java/net/opentsdb/threadpools/UserAwareThreadPoolExecutor.java
```diff
@@ -0,0 +1,352 @@
...
+  protected static final String THREAD_THRESHOLD_PCT = "thread.threshold.pct";
+
+  /**
+   * Percentage of active threads per user to consider before applying the reschedule when the total
+   * percentage of active thread in the executor falls below this threshold value
+   */
+  protected static final String MAX_THREAD_PER_USER_PCT = "max.thread.per.user.pct";
+
+  /** Stores the User thread count that's being current executed */
+  private static final ConcurrentHashMap<String, AtomicInteger> CURRENT_EXECUTIONS =
```
manolama: No statics please. We can just have it as a member of the plugin.

common/src/main/java/net/opentsdb/threadpools/UserAwareThreadPoolExecutor.java
```diff
@@ -0,0 +1,352 @@
...
+        if (COUNT_TO_PURGE.decrementAndGet() <= 0) {
+          CURRENT_EXECUTIONS.entrySet().removeIf(m -> m.getValue().get() == 0);
+          COUNT_TO_PURGE.set(PURGE_CNT);
+        }
+      }
+
+    }
+  }
+
+  class QCFutureWrapper<V> extends FutureTask<V> {
```
manolama: Duplicate code with the QCRunnable. Any way to merge them?

common/src/main/java/net/opentsdb/threadpools/UserAwareThreadPoolExecutor.java
```diff
@@ -0,0 +1,352 @@
...
+      try {
+        // Safe to schedule the task
+        r.run();
+      } finally {
+        // Decrement the counter per user..
+        ai.decrementAndGet();
+
+        // Remove the entry from the state to make sure we don't hold the state for all the users in
+        // the lifetime of the process
+        if (COUNT_TO_PURGE.decrementAndGet() <= 0) {
```
manolama: nice

common/src/main/java/net/opentsdb/threadpools/UserAwareThreadPoolExecutor.java
gskchaitanya: Also, to answer you other question about what happens when the queue is full, the task gets executed in this handler and will run in the same thread that tries to run but rejects it since the queue is full. I'll log it when this happens.

common/src/main/java/net/opentsdb/threadpools/UserAwareThreadPoolExecutor.java
gskchaitanya: Oh,  you think we will have authstate but not user?

common/src/main/java/net/opentsdb/threadpools/UserAwareThreadPoolExecutor.java
gskchaitanya: rgr

common/src/main/java/net/opentsdb/threadpools/UserAwareThreadPoolExecutor.java
manolama: yeah if auth is disabled for any reason we can get null users. There's a default user in that case.

PR #1687, opened 2019-06-25
CORE:

manolama:
- Add the rate to count flag for the Rate config. This is useful for systems like
  StatsD that report the rate per second allowing users to convert back number of
  measurements per interval.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/processor/rate/RateNumericIterator.java
+++ b/core/src/main/java/net/opentsdb/query/processor/rate/RateNumericIterator.java
@@ -41,6 +41,7 @@ import net.opentsdb.query.pojo.RateOptions;
  * @since 3.0
  */
 public class RateNumericIterator implements QueryIterator {
+  private static final long TO_NANOS = 1000000000L;
   
   /** A sequence of data points to compute rates. */
   private final Iterator<TimeSeriesValue<?>> source;
@@ -126,7 +127,6 @@ public class RateNumericIterator implements QueryIterator {
   private void populateNextRate() {
     has_next = false;
     
-    
     while (source.hasNext()) {
       final TimeSeriesValue<NumericType> next = 
           (TimeSeriesValue<NumericType>) source.next();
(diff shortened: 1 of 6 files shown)


core/src/main/java/net/opentsdb/query/processor/rate/RateNumericIterator.java
```diff
@@ -144,6 +143,37 @@ private void populateNextRate() {
...
+      }
+      
+      long prev_epoch = prev_data.timestamp().epoch();
+      long prev_nanos = prev_data.timestamp().nanos();
+      
+      long next_epoch = next.timestamp().epoch();
+      long next_nanos = next.timestamp().nanos();
+      
+      if (next_nanos < prev_nanos) {
+        next_nanos *= 1000000000L;
```
sidhhu: export `1000000000L` to a variable

core/src/test/java/net/opentsdb/query/processor/rate/TestRateConfig.java
```diff
@@ -66,31 +65,35 @@ public void serdes() throws Exception {
...
     assertTrue(options.getDeltaOnly());
+    assertTrue(options.getRateToCount());
   }
   
   @Test
   public void build() throws Exception {
     final RateConfig options = (RateConfig) RateConfig.newBuilder()
         .setCounter(true)
         .setInterval("60s")
+        .setRateToCount(true)
```
sidhhu: also have a test case with interval `auto` please? want to check how that looks like,

core/src/main/java/net/opentsdb/query/processor/rate/RateNumericIterator.java
```diff
@@ -144,6 +143,37 @@ private void populateNextRate() {
...
+        next_epoch--;
+      }
+      
+      long diff = ((next_epoch - prev_epoch) * 1000000000L) + (next_nanos - prev_nanos);
+      double time_delta = (double) diff / (double) config.duration().toNanos();
```
sidhhu: trying to wrap my head around this logic

core/src/test/java/net/opentsdb/query/processor/rate/TestRateConfig.java
manolama: auto isn't used.

PR #1697, opened 2019-07-02
Fix non group_by queries in UI

eswdd:
Ensure we always add the `{}` for group_by filters, otherwise the non group_by filters act as group_by filters


eswdd:
@manolama any chance you could take a look at this? hopefully is pretty simple..

johann8384:
Merged to next.

PR #1701, opened 2019-07-12
3x tweaks

manolama:


core/src/main/java/net/opentsdb/query/filter/TagValueLiteralOrFactory.java
```diff
@@ -44,11 +43,23 @@ public QueryFilter parse(final TSDB tsdb,
     if (node == null) {
       throw new IllegalArgumentException("Node cannot be null.");
     }
-    try {
-      return (QueryFilter) mapper.treeToValue(node, TagValueLiteralOrFilter.class);
-    } catch (JsonProcessingException e) {
-      throw new IllegalArgumentException("Failed to parse TagValueLiteralOr", e);
+    TagValueLiteralOrFilter.Builder builder = TagValueLiteralOrFilter.newBuilder();
+    JsonNode n = node.get("tagKey");
```
smrutilal2: shouldn't this be `node.get("key")`

core/src/main/java/net/opentsdb/query/filter/TagValueLiteralOrFilter.java
```diff
@@ -179,12 +179,12 @@ public static Builder newBuilder() {
   @JsonIgnoreProperties(ignoreUnknown = true)
   public static class Builder {
     @JsonProperty
-    private String tagKey;
+    private String key;
```
smrutilal2: I like this change.

core/src/main/java/net/opentsdb/query/filter/TagValueRangeFilterFactory.java
```diff
@@ -49,11 +48,23 @@ public QueryFilter parse(final TSDB tsdb,
     if (node == null) {
       throw new IllegalArgumentException("Node cannot be null.");
     }
-    try {
-      return (QueryFilter) mapper.treeToValue(node, TagValueRangeFilter.class);
-    } catch (JsonProcessingException e) {
-      throw new IllegalArgumentException("Failed to parse TagValueRegexFilter", e);
+    TagValueRangeFilter.Builder builder = TagValueRangeFilter.newBuilder();
+    JsonNode n = node.get("tagKey");
```
smrutilal2: shouldn't this be node.get("key")

core/src/main/java/net/opentsdb/query/filter/TagValueLiteralOrFactory.java
smrutilal2: Is this for backward compatibility?

core/src/main/java/net/opentsdb/query/filter/TagValueRangeFilterFactory.java
manolama: backwards compatibility. See three lines below.

core/src/main/java/net/opentsdb/query/filter/TagValueLiteralOrFactory.java
manolama: yup

PR #1703, opened 2019-07-15
add free text filter

sidhhu:


core/src/main/java/net/opentsdb/query/filter/FreeTextStringFilter.java
```diff
@@ -0,0 +1,115 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.google.common.base.Objects;
+import com.google.common.base.Strings;
+import com.google.common.hash.HashCode;
+import com.stumbleupon.async.Deferred;
+import net.opentsdb.core.Const;
+import net.opentsdb.stats.Span;
+
+@JsonInclude(Include.NON_NULL)
+@JsonDeserialize(builder = FreeTextStringFilter.Builder.class)
```
manolama: Call me `PassThroughFilter`

PR #1715, opened 2019-07-18
CORE: Write additional props

sidhhu:


sidhhu:
> might wanna change it to `properties` just to save some bytes.

but these are additional. i just wanted it to be consistent with the store.

PR #1723, opened 2019-07-26
CORE:

manolama:
- Fix the ArraySumFactory where it would return a 0 even if we only had a single
  time series with a NaN in it.


smrutilal2:
we should also fix it for other aggregators like min, max etc

PR #1728, opened 2019-07-30
COMMON:

manolama:
- Change the TimeSeriesDataSourceFactory so that it takes a QueryPipelineContext
  in the supportsQuery() call so that method can log to the query.
ALL:
- Support the new change.
- Tweak some casting in the iterator and Java Optional call that Eclipse doesn't like.

### DIFF
--- a/common/src/main/java/net/opentsdb/data/TimeSeriesDataSourceFactory.java
+++ b/common/src/main/java/net/opentsdb/data/TimeSeriesDataSourceFactory.java
@@ -24,7 +24,6 @@ import net.opentsdb.query.QueryNodeConfig;
 import net.opentsdb.query.QueryNodeFactory;
 import net.opentsdb.query.QueryPipelineContext;
 import net.opentsdb.query.TimeSeriesDataSourceConfig;
-import net.opentsdb.query.TimeSeriesQuery;
 import net.opentsdb.query.plan.QueryPlanner;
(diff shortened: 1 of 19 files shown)


core/src/main/java/net/opentsdb/query/processor/topn/TopNResult.java
```diff
@@ -97,7 +97,7 @@ public void run() {
         
         if (value.isInteger()) {
           sorted_results.computeIfAbsent(value.toDouble(), k -> new ArrayList<>()).add(ts);
-        } else {
+        } else if (!Double.isNaN(value.doubleValue())){
```
sidhhu: ah, good catch

common/src/main/java/net/opentsdb/data/TimeSeriesDataSourceFactory.java
```diff
@@ -56,7 +56,7 @@
    * may be empty) or false if the source does not support the query, e.g.
    * maybe the source doesn't handle the particular namespace or metric.
    */
-  public boolean supportsQuery(final TimeSeriesQuery query, 
+  public boolean supportsQuery(final QueryPipelineContext query,
```
sidhhu: rename `query` to `context` please

manolama: Doh.

PR #1733, opened 2019-08-02
rename parentId

sidhhu:


core/src/main/java/net/opentsdb/data/types/event/EventsValue.java
```diff
@@ -85,8 +85,8 @@ protected EventsValue(final Builder builder) {
     this.priority = builder.priority;
     this.userId = builder.userId;
     this.ongoing = builder.ongoing;
-    this.parentId = builder.parentId;
-    this.childId = builder.childId;
+    this.parentId = builder.parentIds;
```
sergeykhegay: Do you want to rename the fields, too?

PR #1739, opened 2019-08-09
Dockerfile that works without a script

itamarst:
The current Dockerfile setup isn't ideal:

1. It's not in the standard location users expect (root directory).
2. It requires a script.

This PR changes so you don't need a script, you just use it like any other Dockerfile.


itamarst:
@johann8384 thanks for reviewing—can you also merge the PR (since I can't) or does someone else have to merge it?

PR #1742, opened 2019-08-12
Status API

### LINKED ISSUE #1584: Feature Request: HTTP /api/status Rest API
Feature Request to add a diagnostic `/api/status` API endpoint.

This should do self diagnostic checks and return an overall `{"status": "ok"}` or something similar (there can be other more detailed fields too).

This is useful for load balanced OpenTSDB farms like I run at scale or OpenTSDB on Kubernetes health checks (a setup I've done recently) as a way of detecting which TSDs are completely healthy and should be in the service endpoint pool.

Currently I use `/api/version` to check the systems are up but this is too rudimentary in my opinion as it isn't really tied to any self diagnostics. Other technologies I've worked with often provide `/api/status` or similar REST API endpoint for health check purposes.

The environment I'm working in currently historically did a metric query check against a metric as their health check but I advised against this as HBase can have partial data outages upon any RegionServer issue or even RegionServer rolling restarts. We later ended up with a complete unnecessary load balancer 503 outage because one region was temporarily unavailable and caused the health checks to fail for all backend TSDs because they all hit that Region down/in-transition, which is why I promptly switched the health checks to something that tests the TSD instances more than the backend or any particular region.

itamarst:
Fixes #1584.

This is still a sketch, albeit a working one, but I would appreciate some feedback.

* [x] I have tested this manually on single-node-HBase-in-Docker, but not against multi-node HBase where the output would be more interesting, e.g. allow "partial".
* [x] Redo manual test now that is code is refactored.
* [x] Unit testing.
* [ ] Is this OK? If the HBase cluster is _completely_ down, the status API never responds. Which ... is maybe OK, cause that's a clear "nope this isn't usable" signal a client/loadbalancer would have to handle too.

@johann8384 maybe you have a little time too look at this?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -417,6 +417,7 @@ test_SRC := \
 	test/tsd/TestRTPublisher.java	\
 	test/tsd/TestSearchRpc.java	\
 	test/tsd/TestStatsRpc.java \
+	test/tsd/TestStatusRpc.java \
 	test/tsd/TestSuggestRpc.java	\
 	test/tsd/TestTreeRpc.java	\
 	test/tsd/TestUniqueIdRpc.java	\
--- a/Makefile.am
+++ b/Makefile.am
@@ -325,6 +325,7 @@ test_SRC := \
 	test/core/TestTags.java	\
 	test/core/TestTSDB.java	\
 	test/core/TestTSDBAddPoint.java	\
+	test/core/TestTSDBTableAvailability.java	\
 	test/core/TestTsdbQueryDownsample.java	\
 	test/core/TestTsdbQueryDownsampleSalted.java	\
 	test/core/TestTsdbQuery.java	\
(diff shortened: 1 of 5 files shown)


itamarst:
I've added unit tests.

chinausers:
There is an extended question, is there a need for open TSDB cluster management mechanism?

chinausers:
Anyway, it's gratifying to see someone maintaining the library.

johann8384:
I've done a cursory review, and this looks pretty good. I want to run it and kick the tires a little, I will try to do that soon.

itamarst:
Testing from user suggests the status API as implemented becomes unresponsive even in the case of partial availability or network issues. This is... not ideal. Probably too-long retry logic somewhere....

itamarst:
As a starting point probably `findTableRegions` in asynchbase, as exposed in `locateRegions`, would need to not have infinite retries. And then OpenTSDB would need to switch to that newer version of Asychbase.

johann8384:
Hey @itamarst, sorry this has been sitting so long. Are you using this code in prod? How is it going?

itamarst:
This code is not being used in prod, I believe my client is using some workaround.

Issue #1753, opened 2019-08-23
Bug: OpenTSDB returning only partial results due to race condition


muffix:
I have noticed that for a small number of queries I was only getting partial results. Looking at the detailed scanner stats, it seems like the results of a single scanner are missing entirely, despite all scanners completing successfully. In our setup with 8 salt buckets, I was missing almost exactly one eighth of the data points. 

This seems to impact all queries, but only when a race condition occurs.  

Looking at the `SaltScanner` code, the call to `getCount()` of the `CountdownLatch` [here](https://github.com/OpenTSDB/opentsdb/blob/b5133bd365a479cbc22f7f0cda888275575a04b6/src/core/SaltScanner.java#L930) is not thread-safe, but the result is used as the key for the map in which the scanners' results are stored. If then two scanners both decrement the latch and then read the same numner, the second one would overwrite the entire results of the first one, which causes them to miss in the final response. 

Happy to put in a PR to get that fixed.

PR #1754, opened 2019-08-23
Fix concurrent result reporting from scanners

muffix:
Fixes a concurrency bug where scanners report their results into a map and would overwrite each other's results.

The call to `getCount()` of the `CountdownLatch` is [not thread-safe](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CountDownLatch.html#getCount()), but the result is used in the map where results of all scanners are stored, which overwrites the results of an individual scanner in a race condition. 

This makes the counter decrements and gets atomic. It also uses the scanner ID as the key which is guaranteed to be unique. The key itself is disregarded later anyway, so it only needs to be unique.

Resolves: #1753

### DIFF
--- a/src/core/SaltScanner.java
+++ b/src/core/SaltScanner.java
@@ -23,8 +23,8 @@ import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.AbstractMap.SimpleEntry;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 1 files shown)


muffix:
Hi @manolama @johann8384, sorry to be annoying, but can I ask for a review of this one and ideally also #1761, which is a similar fix?

We have tested both changes on our production cluster which serves approx. 1200 queries per second on each of our three query nodes. The p99 of query time increased slightly, but we're happy with the performance.

muffix:
[This blog post](https://medium.com/@SkyscannerEng/bosun-mystery-alerts-1167f3f1e0d3#0b99) has a bit more detail on how this bug affected us. There have been no further phantom alerts since we pulled in this fix.

Issue #1083, opened 2017-10-13
Too many series returned when using rolled-up data and specifying single tag key


overstre:
When using rolled-up data in 2.4RC2, specifying a single tag value directly returns multiple series when 1 series was expected.  Putting tag value inside of regexp returns 1 series for rolled-up data as expected.  Using data from default interval also works as expected.

Test data was sent using the api/rollup endpoint for multiple timestamps.  
```     
   test_data = [{"metric": "sys.cpu.nice",
                      "timestamp": ts,
                      "value": 18,
                      "tags": {"host": "web01",
                               "dc": "lga"},
                      "interval": "1h",
                      "aggregator": "SUM"},
                     {"metric": "sys.cpu.nice",
                      "timestamp": ts,
                      "value": 9,
                      "tags": {"host": "web02",
                               "dc": "lga"},
                      "interval": "1h",
                      "aggregator": "SUM"}]
```

Query was
```
/api/query?start=2d-ago&m=sum:1h-sum-none:sys.cpu.nice{host=web01}
```
response was
```
[{"metric":"sys.cpu.nice","tags":{"host":"web01","_aggregate":"RAW","dc":"lga"},"aggregateTags":[],"dps":{"1507762800":18.0,<truncated>}},{"metric":"sys.cpu.nice","tags":{"host":"web02","_aggregate":"RAW","dc":"lga"},"aggregateTags":[],"dps":{"1507762800":9.0,<truncated}}]
```

manolama:
I'll check this one out before cutting it.

manolama:
Found it and fixed in d342301898ccc3edaee18d711a6d9684007d16b8

PR #1756, opened 2019-08-26
Add unit test for rollup filter fix for #1083

neilfordyce:
I was looking into exactly the same issue as #1083 on a fork before spotting that it was already addressed by d342301898ccc3edaee18d711a6d9684007d16b8.  It seems like reasonably complicated logic and I noticed adding some tests is on the todo list https://github.com/OpenTSDB/opentsdb/blob/master/src/core/TsdbQuery.java#L1419.  Since I wrote the unit test before I noticed the fix, I thought I might as well offer it back.

### DIFF
--- a/test/core/TestTsdbQueryQueries.java
+++ b/test/core/TestTsdbQueryQueries.java
@@ -30,6 +30,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import net.opentsdb.query.filter.TagVLiteralOrFilter;
+import net.opentsdb.rollup.RollupConfig;
 import net.opentsdb.rollup.RollupInterval;
 import org.hbase.async.Bytes;
 import org.hbase.async.FilterList;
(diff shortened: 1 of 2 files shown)


test/core/TestTsdbQueryQueries.java
```diff
@@ -1552,6 +1554,75 @@ public void runRegexpNoMatch() throws Exception {
...
+    sub.setDownsample("1h-sum");
+    sub.setFilters(Arrays.asList(new TagVLiteralOrFilter("host", TAGV_STRING)));
+
+    ts_query.setQueries(Arrays.asList(sub));
+    ts_query.validateAndSetQuery();
+    query.configureFromQuery(ts_query, 0);
+
+    final DataPoints[] dps = query.run();
+    assertEquals(1, dps.length);
+    assertEquals(1, dps[0].aggregatedSize());
```
neilfordyce: The test correctly fails on line 1641 if the fix from d342301898ccc3edaee18d711a6d9684007d16b8 isn't included, because aggregatedSize() would return 2 instead of 1.

Issue #1760, opened 2019-08-27
Bug: Race condition causing query results to be incomplete on cold UID caches


muffix:
This is a similar, but different bug from #1753. 

Whenever the scanners finish fetching rows from HBase, they compact the cells and add the new points to [a list](https://github.com/OpenTSDB/opentsdb/blob/b5133bd365a479cbc22f7f0cda888275575a04b6/src/core/SaltScanner.java#L467). The list is not thread-safe, but accessed by multiple threads at a time. This can lead to concurrent modifications and inconsistent query results.

We observed unstable query results when OpenTSDB had to resolve many UIDs to their string values due to the use of regex filters (code starting [here](https://github.com/OpenTSDB/opentsdb/blob/b5133bd365a479cbc22f7f0cda888275575a04b6/src/core/SaltScanner.java#L642)). Our query was looking at high-cardinality metrics, forcing it to resolve ~65000 UIDs. 
On fresh/empty UID caches, these resolutions trigger the resolution and match callbacks asynchronously and from different threads, which means that `processRow()` is also called concurrently from the callback on different threads, modifying the non-thread-safe `kvs` list.

We were able to reliably reproduce the issue with queries for high-cardinality metrics. Symptoms include `ConcurrentModificationException`s and `Found a key value item that was null` in the logs.

PR #1761, opened 2019-08-27
Synchronise the KVs list for scanner results

muffix:
Synchronises the list that holds the KeyValues that have been produced by the scanner callbacks. The list is accessed from multiple threads at a time and wasn't thread-safe, causing inconsistent results and partial loss of data in the response when the UID cache is cold.
Subsequent queries with a warm cache are less susceptible to this bug.

Relates to: #1753
Resolves: #1760


SeanPMiller:
Fixes like these are difficult to test. After some discussion, we agree that the scenario you describe is a thread-safety issue, and we agree that your fix should address it. I see you've identified other issues in the same class.

PR #1762, opened 2019-08-27
Allow end_time = start_time

oyvindwe:
Allow end_time = start_time to be able to query and delete a single datapoint.


oyvindwe:
I have tested this only against the Bigtable backend, where it works as intended.

johann8384:
Merged into `next`.

PR #1775, opened 2019-09-09
COMMON:

manolama:
- Temp API in TypedTimeSeriesIterator to help speed up queries for
  many time series

CORE:
- Temp sharding of group by queries with shared arrays in
  GroupByNumericArrayIterator.
Bottom up Groupby approach


core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -128,17 +152,38 @@ public GroupByNumericArrayIterator(final QueryNode node,
...
+    valuesCombiner = new double[NUM_THREADS][intervals * 2];
+
+    // init with NaN
+    for (int i = 0; i < NUM_THREADS; i++) {
+      Arrays.fill(valuesCombiner[i], Double.NaN);
+    }
 
     // TODO: Need to check if it makes sense to make this threshold configurable
-    if (sources.size() >= 1000) {
+    if (sources.size() >= NUM_THREADS) {
```
smrutilal2: the `num_thereads` should be configurable with a default value of 8.

manolama: Yeah should be configurable but at this spot in the code it's the # of time series we want to switch to parallizing.

PR #1781, opened 2019-09-15
3x tweaks

manolama:


core/src/main/java/net/opentsdb/query/DefaultQueryContextFilter.java
```diff
@@ -0,0 +1,145 @@
...
+ * TODO - Can cache the fetches from the config class for effeciency.
+ *  
+ * @since 3.0
+ */
+public class DefaultQueryContextFilter extends BaseTSDBPlugin 
+    implements QueryContextFilter {
+  private static final Logger LOG = LoggerFactory.getLogger(
+      DefaultQueryContextFilter.class);
+  
+  private static final TypeReference<Map<String, Map<String, String>>> MAP_OF_MAPS =
```
sidhhu: rename to be a bit more meaningful

core/src/main/java/net/opentsdb/query/DefaultQueryContextFilter.java
```diff
@@ -0,0 +1,145 @@
...
+      new TypeReference<Map<String, Map<String, String>>>() { };
+  private static final TypeReference<
+    Map<String, Map<String, Map<String, String>>>> MAP_OF_MAP_OF_MAPS =
```
sidhhu: same here.

PR #1783, opened 2019-09-17
3x tweaks

manolama:


core/src/main/java/net/opentsdb/query/BaseQueryContext.java
```diff
@@ -337,18 +337,36 @@ protected void log(final LogLevel level, final QueryNode node, final String log)
       return;
     }
     
+    final StringBuilder buf = new StringBuilder(15 + 6 + 32 + log.length())
```
SeanPMiller: These numbers in a vacuum are magical. I know they're meaningful, but, someday, using named constants instead could help with intelligibility.

core/src/main/java/net/opentsdb/query/ReadCacheQueryPipelineContext.java
```diff
@@ -526,7 +526,6 @@ void processResults() {
           return;
         }
       }
-      return;
```
SeanPMiller: Conspicuous. Why continue when we previously did not?

manolama: Cause it turns out we can have a successful query when we have some cache misses :) I had this in there by accident.

core/src/main/java/net/opentsdb/query/BaseQueryContext.java
manolama: Added a note and fixed them. Just estimates of the strings we'll be adding so we pre-build the buffer.

core/src/main/java/net/opentsdb/query/ReadCacheQueryPipelineContext.java
```diff
@@ -917,30 +966,51 @@ protected boolean checkComplete() {
           (context.query().getCacheMode() == CacheMode.NORMAL ||
            context.query().getCacheMode() == CacheMode.WRITEONLY)) {
         if (full_query_context.cacheable()) {
+          full_query_caching.set(true);
```
SeanPMiller: What consumes this new field?

manolama: actually.... nothing, need to remove it.

common/src/main/java/net/opentsdb/query/readcache/QueryReadCache.java
```diff
@@ -99,5 +99,22 @@ public void fetch(final QueryPipelineContext context,
...
+  /**
+   * Remove the entry from the cache.
+   * @param timestamp The Unix epoch timestamp in seconds.
+   * @param key A non-null and non-empty key.
+   * @return A deferred resolving to null on complete.
+   */
+  public Deferred<Void> delete(final int timestamp, final byte[] key);
+  
+  /**
+   * Removes the entries from the cache.
```
SeanPMiller: Clarify that timestamps are matched to keys by index.

common/src/main/java/net/opentsdb/query/readcache/QueryReadCache.java
```diff
@@ -99,5 +99,22 @@ public void fetch(final QueryPipelineContext context,
...
+   * @param timestampsAn array of Unix epoch timestamps in seconds, must be
```
SeanPMiller: timestampsAn <- space here

common/src/main/java/net/opentsdb/query/readcache/QueryReadCache.java
```diff
@@ -99,5 +99,22 @@ public void fetch(final QueryPipelineContext context,
...
+   * the same size as the keys.
+   * @param keys A non-null and non-empty array of non-null and non-empty keys.
+   * @return A deferred resolving to null on complete.
+   */
```
SeanPMiller: Doesn't this check array length and potentially throw?

PR #1789, opened 2019-09-23
streaming downsample

smrutilal2:


core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAverageFactory.java
```diff
@@ -107,13 +107,36 @@ public void accumulate(final long[] values,
       }
     }
 
+    @Override
+    public void accumulate(double value, int idx) {
+      if (Double.isNaN(value)) {
```
manolama: We'll need the long version of the interface for the HBase stage but for this, do a `if (double_accumulator == null) <then allocate and copy from long_accumulator>` check in each `accumulate()` implementation.

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/NumericArrayAggregator.java
```diff
@@ -48,7 +48,14 @@ public void accumulate(final long[] values,
...
    */
   public void accumulate(final double[] values);
-  
+
+  /**
+   * Accumulate a double value at a specific index
+   * @param value
+   * @param index
+   */
+  public void accumulate(final double value, final int index);
```
manolama: We'll need the long version for HBase but can leave it out for now.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupBy.java
```diff
@@ -44,9 +54,13 @@
  * @since 3.0
  */
 public class GroupBy extends AbstractQueryNode {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractQueryNode.class);
   
   /** The config for this group by node. */
   private final GroupByConfig config;
+
+  private DownsampleConfig downsampleConfig;
```
manolama: java doc it as `An optional downsample config for use by the GroupByArrayIterator to size it's array properly when running in parallel.`

core/src/main/java/net/opentsdb/query/processor/groupby/GroupBy.java
```diff
@@ -63,7 +77,30 @@ public GroupBy(final QueryNodeFactory factory,
...
+                  TimeSeriesDataSourceConfig config = (TimeSeriesDataSourceConfig) timeSeriesDataSource.config();
+                  List<QueryNodeConfig> pushDownNodes = config.getPushDownNodes();
+                  for (QueryNodeConfig queryNodeConfig : pushDownNodes) {
+                      if(queryNodeConfig instanceof DownsampleConfig) {
+                        downsampleConfig = (DownsampleConfig) queryNodeConfig;
+                        break;
+                      }
+                  }
+                  break;
+                }
```
manolama: For the HBase case, also check to see if the node is an instance of `Downsample` and grab the config from there.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -125,137 +120,131 @@ public GroupByNumericArrayIterator(final QueryNode node,
...
+    } else {
+      size = downsampleConfig.intervals();
     }
-    
-    logger.info("Sources size {}, Intervals found {}", sources.size(), intervals);
 
-    valuesCombiner = new NumericArrayAggregator[NUM_THREADS];
-    for (int i = 0; i < valuesCombiner.length; i++) {
-      valuesCombiner[i] = factory.newAggregator(((GroupByConfig) node.config()).getInfectiousNan());
+    logger.trace("Group size: " + size);
```
manolama: Yank or surround with `if (LOG.isTraceEnabled())`

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -125,137 +120,131 @@ public GroupByNumericArrayIterator(final QueryNode node,
...
+
+    aggregator =
+        factory.newAggregator(((GroupByConfig) node.config()).getInfectiousNan());
+    if (aggregator == null) {
+      throw new IllegalArgumentException(
+          "No aggregator found of type: " + ((GroupByConfig) node.config()).getAggregator());
     }
 
     // TODO: Need to check if it makes sense to make this threshold configurable
     if (sources.size() >= NUM_THREADS) {
```
manolama: Could probably do some factor of NUM_THREADS like `NUM_THREADS * 16`

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -125,137 +120,131 @@ public GroupByNumericArrayIterator(final QueryNode node,
...
-        }
+        accumulate(source, null);
       }
     }
-    
+  }
 
-    double e = ((double) (DateTime.nanoTime() - st)) / 1000_000;
-    logger.info("Groupby iterator took {}, timeseries count {}", e, sources.size());
+  private NumericArrayAggregator createAggregator(
```
manolama: Need's `final`s and we'll need a flag to determine if we should fill and set doubles or keep longs.

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAverageFactory.java
smrutilal2: The long version of this api should be enough I guess

smrutilal2: I think initializing the array would be a prerequisite for this api.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
smrutilal2: Remember, I commented on GSK's pr to make it configurable. This one is on you @manolama

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAverageFactory.java
manolama: rgr

PR #1808, opened 2019-10-09
CORE:

manolama:
- Skip non-metric queries in the RCQPC.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/ReadCacheQueryPipelineContext.java
+++ b/core/src/main/java/net/opentsdb/query/ReadCacheQueryPipelineContext.java
@@ -152,6 +152,22 @@ public class ReadCacheQueryPipelineContext extends AbstractQueryPipelineContext
         break;
       }
       
+      if (config instanceof TimeSeriesDataSourceConfig) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/net/opentsdb/query/ReadCacheQueryPipelineContext.java
```diff
@@ -152,6 +152,22 @@
         break;
       }
       
+      if (config instanceof TimeSeriesDataSourceConfig) {
+        final List<String> types = ((TimeSeriesDataSourceConfig) config).getTypes();
+        if (types != null && !types.isEmpty()) {
+          for (final String type : types) {
+            if (!type.toUpperCase().equals("METRIC")) {
```
sidhhu: why not equalsIgnoreCase and isn't Types an enum??

manolama: oh yeah, I can do that. No types is a string since we don't know what we'll have in the future (traces, logs, etc.)

sidhhu: Oh nvm!!

PR #1815, opened 2019-10-18
Optimize Downsampling and Groupby nodes processing

gskchaitanya:


core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleFactory.java
```diff
@@ -66,13 +66,16 @@
...
   public static final String AUTO_KEY = "tsd.query.downsample.auto.config";
+  public static final String ARRAY_PROCESS_KEY = "tsd.query.downsample.array.process";
   
   public static final TypeReference<Map<String, String>> AUTO_REF = 
       new TypeReference<Map<String, String>>() { };
   
   /** The auto downsample intervals. */
   private List<Pair<Long, String>> intervals;
   
+  private boolean processAsArrays = true;
```
manolama: Just set this as `true` in the config registration.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleFactory.java
```diff
@@ -138,6 +141,15 @@ public DownsampleConfig parseConfig(final ObjectMapper mapper,
     if (n != null) {
       builder.setTimeZone(n.asText());
     }
+
+    n = node.get("process_as_arrays");
```
manolama: camel case please.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleFactory.java
```diff
@@ -350,6 +362,10 @@ public void update(final String key, final Object value) {
           Collections.sort(intervals, REVERSE_PAIR_CMP);
           LOG.info("Updated auto downsample intervals: " + intervals);
         }
+      } else if (key.equals(ARRAY_PROCESS_KEY)) {
+        if (value != null) {
+          processAsArrays = Boolean.valueOf(value.toString());
```
manolama: You can just cast here `(boolean) value`

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
```diff
@@ -0,0 +1,577 @@
...
+import java.time.temporal.ChronoUnit;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.Optional;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Iterator that downsamples data points using an {@link net.opentsdb.data.Aggregator} following
```
manolama: clean out the copy pasta comments and talk about how it converts please.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
```diff
@@ -0,0 +1,577 @@
...
+
+      if (next.timestamp().compare(Op.GTE, this.interval_start)
+          && next.timestamp().compare(Op.LT, this.interval_end)) {
+
+        if (nextTs == null) {
+          // Use the first valid dp timestamp as the first Timestamp.
+          nextTs = next.timestamp().getCopy();
+          long nextRoundoffTs =
+              next.timestamp().epoch() - (next.timestamp().epoch() % intervalPart) + intervalPart;
+          nextTs.updateEpoch(nextRoundoffTs);
```
manolama: There's a `timestamp.snapToPreviousInterval()` for this if you want it.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
```diff
@@ -0,0 +1,577 @@
...
+        if (next.timestamp().compare(Op.GTE, nextTs)) {
+          // flush previous interval once new interval starts
+          flush = true;
+
+          // find the index to update in the Aggregator
+          index = (int) (((nextTs.epoch() - interval_start.epoch() - intervalPart) / intervalPart)
+              % intervals);
+
+          // get the next timestamp in the downsample
+          long nextTsRoundoff =
```
manolama: And there's a `timestamp.add(config.interval())` to use here that'll save some cycles.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
```diff
@@ -0,0 +1,577 @@
...
+        } else {
+          accumulator.add(next.value().doubleValue());
+        }
+
+        if (flush) {
+
+          // increment interval in the groupby aggregator
+          double v = getAggValue(localDoubleAggs, localLongAggs);
+
+          if (numericArrayAggregator instanceof ArrayCount && this.aggEnum == AggEnum.count) {
```
manolama: Add a TODO about making this generic. I don't like having specific aggregation callouts if at all possible. (even though we handle the sum / count case for avg).

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
```diff
@@ -0,0 +1,577 @@
...
+  /**
+   * Enums for aggregation functions
+   *
+   */
+  protected static enum AggEnum {
+    sum, count, min, max, last, avg;
+  }
+
+  /** Enum identifying the aggregation function */
+  private final AggEnum aggEnum;
```
manolama: We should keep this out and generic. Some queries will be coming in with percentile aggs (even though I haven't implemented those as arrays yet) and more.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
```diff
@@ -0,0 +1,577 @@
...
+        numericArrayAggregator.accumulate(v, index);
+      }
+    }
+
+    return null;
+  }
+
+  private double getAggValue(double[] localDoubleAggs, long[] localLongAggs) {
+    double v = Double.NaN;
+    switch (this.aggEnum) {
```
manolama: This is where should still be able to use the new agg interface for array aggregators.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
```diff
@@ -0,0 +1,577 @@
...
+
+        default:
+          throw new UnsupportedOperationException(
+              "Unsupported aggregator function: " + this.aggEnum.name());
+      }
+    }
+  }
+
+  static class Accumulator {
+    private final double[] doubleValues;
```
manolama: To genericize it we should move this into the aggregators classes and use those interfaces.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -166,19 +167,21 @@ public GroupByNumericArrayIterator(
...
-
-    } else {
-      for (final TimeSeries source : sources) {
+    NumericArrayAggregator[] valuesCombiner = new NumericArrayAggregator[NUM_THREADS];
+    for (int i = 0; i < valuesCombiner.length; i++) {
+      valuesCombiner[i] = createAggregator(node, factory, size);
+    }
+    
+    if (size == 0 || !(sources.iterator()
+        .next() instanceof Downsample.DownsampleResult.DownsampleTimeSeries)) {
```
manolama: NPE here if the sources iterator is null or doesn't have a next. e.g. serdes <-- gb <-- time series source

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
gskchaitanya: We cant use config.interval since the storage iterator would only give the next non null interval datapoint that might not be the immediate next dp.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
gskchaitanya: Incase of sparse data, snapToPreviousInterval will be wrong and it should be the last found dp.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
gskchaitanya: This already talks about the new algo :)

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleFactory.java
gskchaitanya: Was following this format from DownsampleConfig.Builder, may be I should not use this format here, will change..

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
gskchaitanya: good point, will take of it.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
gskchaitanya: The reason I didn't have the interfaces/abstraction is to get rid of the references overhead that come with the interfaces. Enums/classes are lighter than the interfaces and since the Downsample iterators are super computational and memory intensive, tried to make them as simple as possible. Will put a TODO here and we shall profile and decide whether to generalize these Data structures.

manolama: we chatted, TODO is fine to get started.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
manolama: NVM, you were right, this is the time series sources, not graph source.

PR #1834, opened 2019-10-29
CORE:

manolama:
- Add dataInterval as a flag to the RateConfig. This avoids introducing spikes
  in the output when converting a stored rate to the count if data is missing.
  The setting is required for summary and array types but can be calculated
  from the raw type.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/processor/rate/RateNumericArrayIterator.java
+++ b/core/src/main/java/net/opentsdb/query/processor/rate/RateNumericArrayIterator.java
@@ -15,6 +15,7 @@
 package net.opentsdb.query.processor.rate;
 
 import com.google.common.reflect.TypeToken;
+
 import net.opentsdb.data.TimeSeries;
 import net.opentsdb.data.TimeSeriesDataType;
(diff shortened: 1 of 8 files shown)


core/src/main/java/net/opentsdb/query/processor/rate/RateNumericArrayIterator.java
```diff
@@ -95,6 +99,21 @@ public RateNumericArrayIterator(final QueryNode node,
...
+      } else {
+        // we shouldn't be here without a configured data interval as we don't 
+        // know the underlying data reporting interval and we can't infer it. So
+        // no counts for you!
+        data_interval = 1;
+      }
+      if (data_interval <= 0) {
+        data_interval = 1;
+      }
+    }
```
gskchaitanya: Otherwise we should still set data_interval  to 1?

manolama: Right we'll assume the count was actually computed earlier. We shouldn't be here unless we're working on a rolled up result and the user passes in valid values. We could validate this in the config in the future.

PR #1845, opened 2019-11-07
Parallelize the top level nodes iteration

gskchaitanya:


core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
```diff
@@ -102,6 +109,12 @@
   //<set ID, <ts hash, ts wrapper>>
   private Map<String, Map<Long, SeriesWrapper>> partials = Maps.newConcurrentMap();
   private List<byte[]> serialized_results = Lists.newArrayList();
+  
+  /** Used for the parallel stream*/
+  private ForkJoinPool customThreadPool = new ForkJoinPool(8);
```
manolama: need to static it so we don't create it each time.

core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
```diff
@@ -241,23 +254,30 @@ public Object call(final ArrayList<TimeSeriesStringId> ids)
...
+                      spec_start,
+                      spec_end,
+                      result);
+                } catch (Exception e) {
+                  LOG.error("Failed to serialize ts: " + series, e);
+                  throw new QueryExecutionException("Unexpected exception "
+                      + "serializing ts: " + series, 0, e);
+                }
+                if (pair.getValue().types().contains(StatusType.TYPE)
+                    && pair.getValue().id() instanceof BaseTimeSeriesByteId) {
```
manolama: won't happen that it's a byte ID series, always string.
BUT we need to sync on the namespace list.

core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
```diff
@@ -509,84 +529,89 @@ private void serializeSeries(
...
-          if (value == null) {
-            continue;
-          }
-          
-          if (iterator.getType() == NumericType.TYPE) {
-            if (writeNumeric((TimeSeriesValue<NumericType>) value, options, 
-                  iterator, json, result, start, end, wrote_values)) {
-              wrote_values = true;
+
+      lock.writeLock().lock(); // since json is not thread safe and we need to form the json in order
```
manolama: lets do a null check and have the lock null if we're not parallelling.

core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
```diff
@@ -241,23 +254,30 @@ public Object call(final ArrayList<TimeSeriesStringId> ids)
...
+                } catch (Exception e) {
+                  LOG.error("Failed to serialize ts: " + series, e);
+                  throw new QueryExecutionException("Unexpected exception "
+                      + "serializing ts: " + series, 0, e);
+                }
+                if (pair.getValue().types().contains(StatusType.TYPE)
+                    && pair.getValue().id() instanceof BaseTimeSeriesByteId) {
+                  BaseTimeSeriesStringId bid = (BaseTimeSeriesStringId) pair.getValue().id();
+                  namespace.append(bid.namespace());
+                  wasStatus.getAndSet(true);
```
manolama: wasStatus needs to be volatile now.

core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
```diff
@@ -226,12 +239,12 @@ public Object call(final ArrayList<TimeSeriesStringId> ids)
...
-          boolean wasStatus = false;
-          boolean wasEvent = false;
-          String namespace = null;
-          if (opts.getParallelThreshold() > 0 &&
-              result.timeSeries().size() > opts.getParallelThreshold()) {
+          AtomicInteger ai = new AtomicInteger(0);
+          AtomicBoolean wasStatus = new AtomicBoolean(false);
+          StringBuilder namespace = new StringBuilder();
+          Collection<QueryNode> topNExists = result.source().pipelineContext().downstreamOfType(result.source(), TopN.class);
+          if (topNExists == null || topNExists.size() == 0) {
```
manolama: And need threshold limit so we only do this if we have more than X number of time series..

gskchaitanya: We can always run it in parallel mode even if it's just a single timeseries

core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
gskchaitanya: good catch

core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
gskchaitanya: wasStatus is an AtomicBoolean, so don't have to make it volatile

core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
gskchaitanya: incase of sequential, one thread shouldn't have an impact on the locking.

PR #1848, opened 2019-11-08
COMMON: optimized tag lookup

smrutilal2:


common/src/main/java/net/opentsdb/data/TimeSeriesStringId.java
```diff
@@ -65,6 +65,8 @@
    */
   public Map<String, String> tags();
 
+  String getTagValue(String key);
```
manolama: tag public just to be similar and mark final and add java doc about it helping when the underlying structure is not a map.

Issue #1686, opened 2019-06-25
Feature request: Combine results from raw and rollup queries


neilfordyce:
I am working on setting up an OpenTSDB with rollups using version 2.4.   Our use case is largely for near real time alerting (requiring sub one minute latency) but occasionally we want to query several months back on high-ish volume data.  Rollups should offer us a great performance improvement here, so thank you for the work in that area.

We are working on creating a daily batch job to create hourly rollups.  That means there will be a delay of up to one day (plus job runtime) to the rollup data being available.  The delay means that any query served by the rollup table would be missing the latest data which is available in the raw table.

If we were stream processing to produce rollups, we would hit similar issues albeit with a lower latency.  We likely wouldn't want the stream processor sending rollup updates as every new raw data point comes in, so there would be a delay in the stream processor producing data points while the rolled up data point "matures".  That delay would cause a similar issue.

We wondered if it might be a reasonable idea to execute the query with some knowledge of what the latest available timestamps are in the rollup table.  With that knowledge, a rollup query could be split up; populating the majority of data from rollup table, but still fall back to the raw table to create the latest data points, then finally joining both sets of data points together.

Does this seem like a reasonable feature or is it too complicated and we would be better off stream processing to reduce latency and just ignore the latest data point?

johnseekins:
We've found using RPC ingest for rollup tables to be an effective trade-off here. It isn't as heavy as the HTTP endpoint and you get close to realtime ingest for rollups as well as raw data.
We are successful at doing this with Kafka. Basically, our rollups are calculated and pushed to a unique Kafka topic and then written to HBase with the Kakfa-RPC system. Works pretty well.

neilfordyce:
Thanks for the information.  It's great to hear from other people using the rollup features.  We are also using the Kafka RPC plugin.  We don't emit points to the rollup Kafka topic until it's unlikely that there will be any more updates to the rollup data point.  

If we run a rollup query, that delay means we miss the latest data points which are available in the raw table.  I was really wondering if it's worth altering the rollup query behaviour to retrieve those latest data points from the raw table.

johnseekins:
Seems an interesting idea, yeah. Personally, I'd like it. The way downsampling works right now makes it look like you're missing data near "now" on graphs. Which is annoying.

neilfordyce:
Awesome.  Thanks for the feedback.  Hoping to open a PR for that.  Will let you know how it goes.

PR #1853, opened 2019-11-13
Add support for splitting rollup queries

muffix:
Adds a SplitRollupQuery class that suports splitting a rollup query into
two separate queries.
This is useful for when a rollup table is filled by e.g. a batch job
that processes the data from the previous day on a daily basis. Rollup
data for yesterday will then only be available some time today. This
delay SLA can be configured on a per-table basis. The delay would
specify by how much time the table can be behind real time.

If a query comes in that would query data from that blackout period
where data is only available in the raw table, but not yet guaranteed to
be in the rollup table, the incoming query can be split into two using
the SplitRollupQuery class. It wraps a query that queries the rollup
table until the last guaranteed to be available timestamp based on the
SLA; and one that gets the remaining data from the raw table.

Previously, without splitting and a lagging rollup table, the result was
something like this:
image

With splitting and stitching the results, the data is complete:
image

Query splitting needs to be enabled globally by 
- setting `tsd.rollups.split_query.enable = true` in the OpenTSDB config
- defining how far behind real time a rollup table can be in the worst case in the rollup table config, e.g. `"delaySla": "2d"` if, like in the following example, we know that the 1h rollup table is never going to be laging by more than two days:
```javascript
{
    "aggregationIds": {
        "sum": 0,
        "count": 1,
        "min": 2,
        "max": 3,
        "avg": 4
    },
    "intervals": [
        {
            "table": "tsdb",
            "preAggregationTable": "tsdb",
            "interval": "1m",
            "rowSpan": "1h",
            "defaultInterval": true
        },
        {
            "table": "tsdb-rollup-1h",
            "preAggregationTable": "tsdb-rollup-1h",
            "interval": "1h",
            "rowSpan": "1d",
            "delaySla": "2d",
            "defaultInterval": false
        }
    ]
}
```
Everything until now-2d will then be retrieved as raw data points from the table `tsdb` and all points before that from `tsdb-rollup-1h`. Results are combined into a single response so that users don't need to worry about how far rollup data is available. 

Resolves: #1686

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,6 +32,7 @@ dist_noinst_DATA = pom.xml.in build-aux/rpm/opentsdb.conf \
   build-aux/rpm/logback.xml build-aux/rpm/init.d/opentsdb \
   build-aux/rpm/systemd/opentsdb@.service
 tsdb_SRC := \
+	src/core/AbstractQuery.java \
 	src/core/AggregationIterator.java	\
 	src/core/Aggregator.java	\
 	src/core/Aggregators.java	\
--- a/Makefile.am
+++ b/Makefile.am
@@ -48,6 +48,7 @@ tsdb_SRC := \
 	src/core/DownsamplingSpecification.java \
 	src/core/FillingDownsampler.java \
 	src/core/FillPolicy.java \
+	src/core/GroupCallback.java \
 	src/core/Histogram.java	\
 	src/core/HistogramAggregation.java	\
 	src/core/HistogramAggregationIterator.java	\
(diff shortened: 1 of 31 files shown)


johann8384:
@muffix I'd like to merge this, can you confirm you've completed the CLA? https://docs.google.com/spreadsheet/embeddedform?formkey=dFNiOFROLXJBbFBmMkQtb1hNMWhUUnc6MQ

muffix:
@johann8384 I think I've done it a while ago, but wasn't entirely sure. So I resubmitted it now, just in case.

iamsurya-dev:
👏 👏

PR #1855, opened 2019-11-14
CORE:

manolama:
- Add GroupByNumericSummaryParallelIterator to allow paralleling summary
  queries.

ASYNCHBASE:
- Fix a potential NPE for timed out queries in the scanner class.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericSummaryParallelIterator.java
+++ b/core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericSummaryParallelIterator.java
@@ -1,0 +1,354 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2019  The OpenTSDB Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package net.opentsdb.query.processor.groupby;
+
+import java.time.temporal.TemporalAmount;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+
+import com.google.common.reflect.TypeToken;
+
+import net.opentsdb.data.TimeSeries;
+import net.opentsdb.data.TimeSeriesDataType;
+import net.opentsdb.data.TimeSeriesValue;
+import net.opentsdb.data.TimeStamp;
+import net.opentsdb.data.TypedTimeSeriesIterator;
+import net.opentsdb.data.types.numeric.MutableNumericSummaryValue;
+import net.opentsdb.data.types.numeric.NumericSummaryType;
+import net.opentsdb.data.types.numeric.NumericType;
+import net.opentsdb.exceptions.QueryDownstreamException;
+import net.opentsdb.query.QueryIterator;
+import net.opentsdb.query.QueryNode;
+import net.opentsdb.query.QueryResult;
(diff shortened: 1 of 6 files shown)


core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericSummaryParallelIterator.java
```diff
@@ -0,0 +1,355 @@
...
+    DownsampleConfig downsampleConfig = ((GroupBy) node).getDownsampleConfig();
+    if (null == downsampleConfig) {
+      throw new IllegalStateException("Shouldn't be here if the downsample was null.");
+    }
+    intervals = downsampleConfig.intervals();
+    interval = downsampleConfig.interval();
+    interval_in_seconds = DateTime.parseDuration(downsampleConfig.getInterval()) / 1000;
+    start_epoch = (int) downsampleConfig.startTime().epoch();
+    ts = downsampleConfig.startTime().getCopy();
+    System.out.println("        START: " + ts.epoch());
```
gskchaitanya: Remove or move it  to debug

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericSummaryParallelIterator.java
```diff
@@ -0,0 +1,355 @@
...
+    // TODO
+    expect_sums_and_counts = false;
+    
+    DownsampleConfig downsampleConfig = ((GroupBy) node).getDownsampleConfig();
+    if (null == downsampleConfig) {
+      throw new IllegalStateException("Shouldn't be here if the downsample was null.");
+    }
+    intervals = downsampleConfig.intervals();
+    interval = downsampleConfig.interval();
+    interval_in_seconds = DateTime.parseDuration(downsampleConfig.getInterval()) / 1000;
```
gskchaitanya: We can infer this from downsampleConfig.interval() instead of parsing the interval string again.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericSummaryParallelIterator.java
```diff
@@ -0,0 +1,355 @@
...
+    results = new double[intervals];
+    Arrays.fill(results, Double.POSITIVE_INFINITY); // sentinel
+    
+    long[] counts = null;
+    if (agg == AggEnum.avg) {
+      counts = new long[intervals];
+    }
+    for (int i = 0; i < accumulators.length; i++) {
+      switch (agg) {
+      case sum:
```
gskchaitanya: accumulate for sum aswell?

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericSummaryParallelIterator.java
```diff
@@ -0,0 +1,355 @@
...
+          if (Double.isInfinite(results[x])) {
+            results[x] = accumulators[i].accumulator[x];
+          } else {
+            if (accumulators[i].accumulator[x] > results[x]) {
+              results[x] = accumulators[i].accumulator[x];
+            }
+          }
+        }
+        break;
+      case max:
```
gskchaitanya: How about max?

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericSummaryParallelIterator.java
manolama: Can't, tried it out and if you do `interval.get(ChronoUnit.SECONDS)` for a wide time range like 1 month, it throws an unsupported unit exceptions. So we have to double parse it (but eventually we'll move that out.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericSummaryParallelIterator.java
manolama: Good catch

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericSummaryParallelIterator.java
manolama: falls through

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericSummaryParallelIterator.java
manolama: oh wait, it was right the first time. this is for the sum OR count case. Count is handled in avg below if we want the avg.

PR #1857, opened 2019-11-15
CORE:

manolama:
- Turns out we need the max of either the real count or the given minimum
  interval for downsampling.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
+++ b/core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
@@ -358,20 +358,15 @@ public class DownsampleNumericToNumericArrayIterator
 
         break;
       case avg:
-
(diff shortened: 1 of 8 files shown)


core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleNumericToNumericArrayIterator.java
```diff
@@ -358,20 +358,18 @@ private double getAggValue(double[] localDoubleAggs, long[] localLongAggs) {
 
         break;
       case avg:
-
         if (localDoubleAggs[5] == 0 && localLongAggs[5] == 0) {
+          LOG.info("      REPTAVG?: " + (localDoubleAggs[1] + localLongAggs[1]) + "  DPS: " + config.dpsInInterval());
```
gskchaitanya: Remove these :)

manolama: argg!

PR #1859, opened 2019-11-15
use the metric from query for multi gets

sidhhu:


implementation/elasticsearch/src/main/java/net/opentsdb/meta/impl/es/ESMetaResponse.java
```diff
@@ -337,9 +338,9 @@ public ESMetaResponse(final Map<String, MultiSearchResponse> response) {
                 }
               case TIMESERIES:
                 if (result == null) {
-                  result = parseTimeseries(query, meta_query, response, null);
+                  result = parseTimeseries(query, meta_query, response, isMultiGet ? true : false, null);
                 } else {
-                  parseTimeseries(query, meta_query, response, result);
+                  parseTimeseries(query, meta_query, response, isMultiGet ? true : false, result);
```
manolama: isMultiGet is always false here cause of the check above.

PR #1865, opened 2019-11-19
3x tweaks

manolama:


core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -178,10 +178,8 @@ public GroupByNumericArrayIterator(
       this.statsCollector = tsdb.getStatsCollector();
 
       if (this.result.isSourceProcessInParallel()) {
-        logger.debug("Accumulate in parallel, source size {}", sources.size());
```
gskchaitanya: Can we keep this debug log, please? Helps during debugging

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -178,10 +178,8 @@ public GroupByNumericArrayIterator(
...
         accumulateInParallel(sources, valuesCombiner);
       } else {
-        logger.debug("Accumulate in sequence, source size {}", sources.size());
```
gskchaitanya: Also this one ^

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
manolama: How bout trace?

gskchaitanya: We have a deal!

PR #1866, opened 2019-11-19
CORE:

sidhhu:
Handle process in parallel in serde


core/src/main/java/net/opentsdb/query/processor/topn/TopNResult.java
```diff
@@ -144,5 +144,9 @@ public int compareTo(final SortOnDouble o) {
     }
     
   }
-  
+
+  @Override
+  public boolean processInParallel() {
+    return true;
```
manolama: Nope, make it false please :)

PR #1874, opened 2019-11-22
3x tweaks

manolama:


implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringBaseline.java
```diff
@@ -0,0 +1,251 @@
...
+        new com.yahoo.egads.data.TimeSeries();
+    final double[] results = new double[(int) node.predictionIntervals()];
+    
+    // fill the prediction with nans at the proper timestamps
+    long ts = node.predictionStart();
+    for (int i = 0; i < results.length; i++) {
+      try {
+        prediction.append(ts, Float.NaN);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
```
SeanPMiller: Converting checked to unchecked makes me nervous in a language that pushes hard to handle checked. Where do you intend this to be handled?

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
```diff
@@ -0,0 +1,815 @@
...
+      QueryExecutionException e = new QueryExecutionException("Successfully "
+          + "started prediction start [" + prediction_start + "] and key " 
+          + Arrays.toString(cache_key) + " State: " + JSON.serializeToString(state), 423);
+      sendUpstream(e);
+    }
+  }
+  
+  void countdown() {
+    latch.countDown();
+    if (latch.getCount() == 0) {
```
SeanPMiller: The `getCount()` method is not `synchronized` or otherwise synchronized. If you want to use this for control, which it seems you do, then we need a different method.

See: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CountDownLatch.html#getCount--

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
```diff
@@ -0,0 +1,815 @@
...
+  long predictionIntervals() {
+    return prediction_intervals;
+  }
+  
+  long predictionInterval() {
+    return prediction_interval;
+  }
+  
+  void writeCache(final QueryResult result) {
+    if (cache_hit || cache == null) {
```
SeanPMiller: If `CacheCB.call()` is called with a non-`null` value, then it will not branch and will eventually set `prediction` to that argument, set `cache_hit` to `true`, and count down the latch. If it is possible for another thread to have entered `writeCache()` concurrently, then there is a data race on `cache_hit`. The `volatile` modifier itself does not resolve races.

There may be no problem here. I need more time to analyze program flow.

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
```diff
@@ -0,0 +1,815 @@
...
+
+  @Override
+  public void close() {
+    // TODO fire out out what to do here. For now, we'll let the baseline keep
+    // going.
+  }
+
+  @Override
+  public void onNext(final QueryResult next) {
+    if (data_source == null || !data_source.equals(next.dataSource())) {
```
SeanPMiller: I wonder if this will be compiled as one or two `volatile` reads.

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
```diff
@@ -0,0 +1,815 @@
...
+      synchronized (this) {
+        if (data_source == null || !data_source.equals(next.dataSource())) {
+          data_source = next.dataSource();
```
SeanPMiller: This is a DCLP-protected write.

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
```diff
@@ -0,0 +1,815 @@
...
+
+    @Override
+    public void onNext(final QueryResult next) {
+      updateState(State.RUNNING, null);
+      // TODO filter, for now assume one result
+      
+      if (data_source == null) {
+        synchronized (OlympicScoringNode.this) {
+          if (data_source == null) {
+            data_source = next.dataSource();
```
SeanPMiller: This DCLP enforces that one and only one thread ever writes to the `data_source` when it is `null`.

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
```diff
@@ -0,0 +1,815 @@
...
+  @Override
+  public void onNext(final QueryResult next) {
+    if (data_source == null || !data_source.equals(next.dataSource())) {
+      synchronized (this) {
+        if (data_source == null || !data_source.equals(next.dataSource())) {
+          data_source = next.dataSource();
+        }
+      }
+    }
+    current = next;
```
SeanPMiller: Meanwhile, this `volatile` write synchronizes with nothing. Should it?

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringBaseline.java
manolama: These should be handled by the caller. I really wish EGADs didn't have the checked exceptions and I may yank them out.

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
manolama: Yeah I can switch to an atomic int instead, I always start with these just so I know what's going on when roughing it out.

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
manolama: Yeah I really just have volatile to make sure we don't cache the pointer. Only one path and one thread will ever call into this function. I just can't guarantee which thread that will be.

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
manolama: yeah, it breaks the typically DCLP but I'm not worried too much as the values *should* always be the same. I could correct it with a lock and actually throw an error if the data sources are different.

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
manolama: no, it will only be called once (at least it will after I address the filtering issue in a later PR)

Issue #846, opened 2016-08-15
Support for microsecond precision


IceMan81:
Is support for microsecond precision planned? There are a lot of scenarios where microsecond support is critical like a time series of stock prices or other market events. Not storing the data points with its original precision is a significant data loss. Reality is that there are real world use cases where there is data or multiple data point every microsecond or sub-microsecond.

manolama:
We've been thinking about it and it's a fair amount of work to go down to the us level. We may be able to get it into v3.

IceMan81:
Its probably best to design for even greater precision than microseconds. Many timeseries dbs are heading in this directions because of requirements to maintain upto nanosecond precision. 
I'm also hoping to looking into the details of the OpenTSDB design and see what needs to be done.

girichills:
Support for nano seconds is much needed as otherwise you are restricting the scope and usage of the product itself  . There are a lot of use cases wherein very high frequency data should be captured and with the emergence of technologies like IOT , such requirements will only grow multi folds .

Explorer89:
@manolama , Is this feature considered for 3.0

timhughes:
Microsecond and lower resolution is now coming into force in the European Finance industry due to MiFID2. Having a TSDB that is capable of this resolution that can record massive amounts of data is rapidly becoming essential in the industry. This would be a great feature to support. I am currently looking for something suitable but the millisecond resolution of OpenTSDB means it is currently out of the running.

Explorer89:
Any work is done on this ?

Issue #845, opened 2016-08-12
Endpoint /api/query/exp not responding when groupBy attribute is false 


aemitos:
Hi guys, we are trying to use expressions with the endpoint /api/query/exp.
We are sending this expression below:

```
{
   "time": {
       "start": "2015/01/01",
       "end":"2015/01/04",
       "aggregator":"sum"
   },
   "filters": [
       {
           "tags": [
               {
                   "type": "iliteral_or",
                   "tagk": "key",
                   "filter": "0|1",
                   "groupBy":false
               }
           ],
           "id": "f1"
       },{
           "tags": [
               {
                   "type": "literal_or",
                   "tagk": "key",
                   "filter": "0",
                   "groupBy": false
               }
           ],
           "id": "f2"
       }
   ],
   "metrics": [
       {
           "id": "a",
           "metric": "test_put1",
           "filter": "f1",
           "fillPolicy":{"policy":"nan"}
       },
       {
           "id": "b",
           "metric": "test_put2",
           "filter": "f2",
           "fillPolicy":{"policy":"nan"}
       }
   ],
   "expressions": [
       {
           "id": "c",
           "expr": "a + b"
       }
    ],
    "outputs":[
      {"id":"c", "alias":"Test expression"}
    ]
 }
```
If we change the filter's flag "groupBy" to `true`, the expression works fine but the result per point is `[1420077600000,3500,100]`, like below:
```
{
  "outputs": [
    {
      "id": "c",
      "alias": "Test expression",
      "dps": [
        [
          1420077600000,
          3500,
          100
        ],
        [
          1420164000000,
          3500,
          100
        ],
        [
          1420250400000,
          3500,
          100
        ],
        [
          1420336800000,
          3500,
          100
        ]
      ],
      "dpsMeta": {
        "firstTimestamp": 1420077600000,
        "lastTimestamp": 1420336800000,
        "setCount": 4,
        "series": 2
      },
      "meta": [
        {
          "index": 0,
          "metrics": [
            "timestamp"
          ]
        },
        {
          "index": 1,
          "metrics": [
            "test_put1",
            "test_put2"
          ],
          "commonTags": {
            "key": "0"
          },
          "aggregatedTags": []
        },
        {
          "index": 2,
          "metrics": [
            "test_put1",
            "test_put2"
          ],
          "commonTags": {
            "key": "1"
          },
          "aggregatedTags": []
        }
      ]
    }
  ],
  "query": {
    "name": null,
[...]

PR #1881, opened 2019-12-03
CORE:

manolama:
- Add the ability to store deltas in the anomaly prediction as a config
  flag.

EGADS:
- Compute the deltas.
- Add a cas check to the state updater

REDIS:
- Use setex to properly update the state.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/router/TimeRouterFactory.java
+++ b/core/src/main/java/net/opentsdb/query/router/TimeRouterFactory.java
@@ -115,7 +115,7 @@ public class TimeRouterFactory extends BaseTSDBPlugin implements
     if (context.query().isDebugEnabled()) {
       StringBuilder buf = new StringBuilder();
       for (int i = 0; i < sources.size(); i++) {
-        if (i >= 0) {
+        if (i >= 0 && buf.length() > 0) {
           buf.append(", ");
         }
         buf.append(sources.get(i).getSourceId());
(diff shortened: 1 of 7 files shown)


core/src/main/java/net/opentsdb/query/router/TimeRouterFactory.java
```diff
@@ -115,7 +115,7 @@ public void setupGraph(final QueryPipelineContext context,
     if (context.query().isDebugEnabled()) {
       StringBuilder buf = new StringBuilder();
       for (int i = 0; i < sources.size(); i++) {
-        if (i >= 0) {
+        if (i >= 0 && buf.length() > 0) {
```
gskchaitanya: `i > 0` should be enough here

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
```diff
@@ -797,12 +821,35 @@ void updateState(final State new_state, final Exception e) {
...
+      try {
+        Thread.sleep(50);
+      } catch (InterruptedException e1) {
+        LOG.error("Unexpected interruption", e1);
+        return;
+      }
+      cache.setState(cache_key, state, 300_000); // TODO config
+      cas = cache.getState(cache_key);
+    }
+    if (cas.equals(state)) {
```
gskchaitanya: If another thread updates the cache just  after #845  and before #846, `cas` won't be equal to `state`. If we are just logging, we should be ok.

core/src/main/java/net/opentsdb/query/router/TimeRouterFactory.java
manolama: should but I had a funky empty string popping up.

implementation/egads/src/main/java/net/opentsdb/query/anomaly/egads/olympicscoring/OlympicScoringNode.java
manolama: yeah just logging

PR #1884, opened 2019-12-04
EXECUTORS::

sidhhu:
Fix the process in parallel flag for events

### DIFF
--- a/executors/http/src/main/java/net/opentsdb/query/execution/HttpQueryV3Result.java
+++ b/executors/http/src/main/java/net/opentsdb/query/execution/HttpQueryV3Result.java
@@ -90,6 +90,9 @@ public class HttpQueryV3Result implements QueryResult {
 
   /** Object Mapper for serdes. */
   private final ObjectMapper mapper = new ObjectMapper();
+
+  /** Flag to know if we need to process series parallely */
+  private boolean process_parallel = true;
   
   /**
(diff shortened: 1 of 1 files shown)


executors/http/src/main/java/net/opentsdb/query/execution/HttpQueryV3Result.java
```diff
@@ -90,6 +90,9 @@
 
   /** Object Mapper for serdes. */
   private final ObjectMapper mapper = new ObjectMapper();
+
+  /** flag to know if we need to process series parallely */
+  private boolean process_parallel = true;
```
gskchaitanya: Make this volatile so the serde threads can read right values

manolama: doesn't need to be, it'll only be read from one thread.

gskchaitanya: Serde can now process the groupby units parallelly if needed, maybe this Query Result type won't be processed by Serde?

sidhhu: It will be, but they look at this flag and take a call to process parallelly or not. This call should be from one thread always.

PR #1889, opened 2019-12-16
CORE:

smrutilal2:
- interleaving big and small queries to minimize starvation

### DIFF
--- a/common/src/main/java/net/opentsdb/utils/TSDBQueryQueue.java
+++ b/common/src/main/java/net/opentsdb/utils/TSDBQueryQueue.java
@@ -1,3 +1,17 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2015-2017  The OpenTSDB Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 package net.opentsdb.utils;
 
 public interface TSDBQueryQueue<T> {
--- a/common/src/main/java/net/opentsdb/utils/TSDBQueryQueue.java
+++ b/common/src/main/java/net/opentsdb/utils/TSDBQueryQueue.java
@@ -1,0 +1,12 @@
+package net.opentsdb.utils;
+
+public interface TSDBQueryQueue<T> {
+
+  void put(T t);
+
+  T take() throws InterruptedException;
+
+  int size();
+
+  void shutdown();
+}
--- a/core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
+++ b/core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
@@ -93,7 +93,7 @@ public class JsonV3QuerySerdes implements TimeSeriesSerdes {
 
   /** The options for this serialization. */
   private final SerdesOptions options;
-  
+
   /** The query start and end timestamps. */
   private final TimeStamp start;
   private final TimeStamp end;
@@ -108,7 +108,9 @@ public class JsonV3QuerySerdes implements TimeSeriesSerdes {
   //<set ID, <ts hash, ts wrapper>>
   private Map<String, Map<Long, SeriesWrapper>> partials = Maps.newConcurrentMap();
   private List<byte[]> serialized_results = Lists.newArrayList();
-  
+
+  /** Used for the parallel stream*/
+  private static ForkJoinPool customThreadPool = new ForkJoinPool(8);
   /** Lock to parallelize the iterator*/
   private ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
 
@@ -131,7 +133,7 @@ public class JsonV3QuerySerdes implements TimeSeriesSerdes {
     this.context = context;
     this.options = options;
     this.stream = stream;
-    
+
     start = context.query().startTime();
     end = context.query().endTime();
   }
@@ -158,7 +160,7 @@ public class JsonV3QuerySerdes implements TimeSeriesSerdes {
       throw new RuntimeException("Failed to instantiate a JSON "
           + "generator", e1);
     }
-    
+
     final List<TimeSeries> series;
     final List<Deferred<TimeSeriesStringId>> deferreds;
     if (result.idType() == Const.TS_BYTE_ID) {
@@ -180,14 +182,14 @@ public class JsonV3QuerySerdes implements TimeSeriesSerdes {
 
       @Override
       public Object call(final ArrayList<TimeSeriesStringId> ids)
-            throws Exception {
+          throws Exception {
         try {
           json.writeStartObject();
-          json.writeStringField("source", result.source().config().getId() 
+          json.writeStringField("source", result.source().config().getId()
               + ":" + result.dataSource());
           final TimeStamp spec_start;
           final TimeStamp spec_end;
-          
+
           // serdes time spec if present
           if (result.timeSpecification() != null) {
             if (result.timeSpecification().start().compare(Op.LT, start)) {
@@ -204,7 +206,7 @@ public class JsonV3QuerySerdes implements TimeSeriesSerdes {
               spec_start = result.timeSpecification().start();
             }
 
-             if (result.timeSpecification().end().compare(Op.GT, end)) {
+            if (result.timeSpecification().end().compare(Op.GT, end)) {
               spec_end = end.getCopy();
               int interval = DateTime.getDurationInterval(
                   result.timeSpecification().stringInterval());
@@ -214,7 +216,7 @@ public class JsonV3QuerySerdes implements TimeSeriesSerdes {
             } else {
               spec_end = result.timeSpecification().end();
             }
-             
+
             json.writeObjectFieldStart("timeSpecification");
             // TODO - ms, second, nanos, etc
             json.writeNumberField("start", spec_start.epoch());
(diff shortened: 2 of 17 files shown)


core/src/main/java/net/opentsdb/query/processor/groupby/BigSmallLinkedBlockingQueue.java
```diff
@@ -0,0 +1,91 @@
+package net.opentsdb.query.processor.groupby;
+
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Predicate;
+
+public class BigSmallLinkedBlockingQueue<T> {
```
manolama: javadoc about what this is intended to do.

core/src/main/java/net/opentsdb/query/processor/groupby/BigSmallLinkedBlockingQueue.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  public void put(T t) {
+    boolean isBig = bigPredicate.test(t);
+    ConcurrentLinkedQueue<T> q = isBig ? bigQ : smallQ;
+
+    q.offer(t);
+    int c = count.getAndIncrement();
+
+    if (c == 0) {
```
manolama: `>`

core/src/main/java/net/opentsdb/query/processor/groupby/BigSmallLinkedBlockingQueue.java
```diff
@@ -0,0 +1,91 @@
...
+  public int size() {
+    return count.get();
+  }
+
+  public void put(T t) {
+    boolean isBig = bigPredicate.test(t);
+    ConcurrentLinkedQueue<T> q = isBig ? bigQ : smallQ;
+
+    q.offer(t);
+    int c = count.getAndIncrement();
```
manolama: And `incrementAndGet` to set it properly for below.

core/src/main/java/net/opentsdb/query/processor/groupby/BigSmallLinkedBlockingQueue.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public T take() throws InterruptedException {
+    T next;
+    try {
+      readLock.lockInterruptibly();
+      while (count.get() == 0) {
+        readSignal.await();
+      }
+      next = doPoll();
+      int c = count.getAndDecrement();
```
manolama: `decrementAndGet()`

core/src/main/java/net/opentsdb/query/processor/groupby/BigSmallLinkedBlockingQueue.java
```diff
@@ -0,0 +1,91 @@
...
+      if (c > 1) {
```
manolama: `> 0`

core/src/main/java/net/opentsdb/query/processor/groupby/BigSmallLinkedBlockingQueue.java
```diff
@@ -0,0 +1,91 @@
...
+   * the synchronized construct
+   *
+   * @return head of the queue
+   */
+  private T doPoll() {
+    ConcurrentLinkedQueue<T> nextQ = takeFromBigQ ? bigQ : smallQ;
+    T next = nextQ.poll();
+    if (next == null) {
+      ConcurrentLinkedQueue<T> otherQ = takeFromBigQ ? smallQ : bigQ;
+      next = otherQ.poll();
```
manolama: If this is null as well should we just return null here and not flip the toggle below?

core/src/main/java/net/opentsdb/query/processor/groupby/GroupBy.java
```diff
@@ -68,6 +65,8 @@
    */
   private DownsampleConfig downsampleConfig;
 
+  private int timeSeriesCount;
```
manolama: Unfortunately we can't store this here as a GB may get multiple results (depend on how the user configures their graph). Instead we need to pass the Result into the GB result as a source result, then let the GB Iterators read from that.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -77,10 +78,78 @@
...
+        CountDownLatch doneSignal,
+        StatsCollector statsCollector) {
+      this.totalTsCount = totalTsCount;
+      this.tsList = tsList;
+      this.startIndex = startIndex;
+      this.endIndex = endIndex;
+      this.combiner = combiner;
+      this.doneSignal = doneSignal;
+      this.statsCollector = statsCollector;
+      statsCollector.setGauge("groupby.timeseries.count", tsList.size());
```
manolama: Probably want a counter here so we can track average, otherwise we just get the high-watermark right?

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -77,10 +78,78 @@
...
+    @Override
+    public void run() {
+      statsCollector.addTime("groupby.queue.wait.time", System.nanoTime() - s, ChronoUnit.NANOS);
+      for (int i = startIndex; i < endIndex; i++) {
+        doRun(tsList.get(i), combiner);
+      }
+      doneSignal.countDown();
+    }
+
+    public abstract void doRun(TimeSeries timeSeries, Combiner combiner);
```
manolama: javadoc me

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -77,10 +78,78 @@
...
+  }
+
+  protected static BigSmallLinkedBlockingQueue<GroupByJob> blockingQueue =
```
manolama: And actually could you move this as a member of the `GroupByFactory` please? I hate statics. And we can add a thread there to periodically report the size of the queue.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -77,10 +78,78 @@
...
+          try {
+            blockingQueue.take().run();
+          } catch (InterruptedException ignored) {
+            logger.error("GroupBy thread interrupted", ignored);
+          }catch (Throwable throwable) {
+            logger.error("Error running GroupBy job", throwable);
+          }
+        }
+      }, "Group by thread: " + (i + 1));
+      thread.setDaemon(true);
```
manolama: won't need if part of the GBF.

manolama: Also name the thread please. Nicer in the thread dumps.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -267,14 +336,19 @@ private void accumulateInParallel(
       final List<TimeSeries> tsList, final NumericArrayAggregator[] combiners) {
 
     final int tsCount = tsList.size();
-    final int jobCount = combiners.length;
-    final int tsPerJob = tsCount / jobCount;
-    final List<Future<?>> futures = new ArrayList<>(jobCount);
+    final int threadCount = combiners.length;
+    int tsPerJob = tsCount / threadCount;
+    if(tsPerJob > MAX_TS_PER_JOB) {
```
manolama: Make this a config in the GroupByFactory please.

core/src/test/java/net/opentsdb/query/processor/groupby/TestBigSmallLinkedBlockingQueue.java
```diff
@@ -0,0 +1,171 @@
+package net.opentsdb.query.processor.groupby;
```
manolama: LICENSE header

core/src/main/java/net/opentsdb/query/processor/groupby/BigSmallLinkedBlockingQueue.java
manolama: or just delete the check as we always want to notify a thread when someone writes.

core/src/main/java/net/opentsdb/query/processor/groupby/BigSmallLinkedBlockingQueue.java
smrutilal2: I guess, it's okay to toggle. At least the next read will check the other q.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
smrutilal2: it's covered already. `guage` reports `sum`, `count`, `min`, `max` and `lastValue`.

May be we need a break up for both the `big` and `small` q.

common/src/main/java/net/opentsdb/utils/TSDBQueryQueue.java
```diff
@@ -0,0 +1,12 @@
+package net.opentsdb.utils;
```
manolama: License header

common/src/main/java/net/opentsdb/utils/TSDBQueryQueue.java
```diff
@@ -0,0 +1,12 @@
...
+
+public interface TSDBQueryQueue<T> {
```
manolama: javadocs for all of these.

core/src/main/java/net/opentsdb/query/execution/serdes/JsonV3QuerySerdes.java
```diff
@@ -244,40 +242,33 @@ public Object call(final ArrayList<TimeSeriesStringId> ids)
           if (result.processInParallel()) {
             List<TimeSeries> tss = result.timeSeries();
             LOG.debug("Processing the iterators parallelly: " + tss.size());
-            final List<Pair<Integer, TimeSeries>> pairs =
-                Lists.newArrayListWithExpectedSize(tss.size());
-            idx = 0;
             for (int i = 0; i < tss.size(); i++) {
               TimeSeries ts = tss.get(i);
-              pairs.add(new Pair<Integer, TimeSeries>(idx++, ts));
+              try {
```
manolama: This is the parallel processing path so this would revert to serialization.

manolama: Let's leave this as is for now and we'll revisit it.

PR #1896, opened 2019-12-27
Improve ExplicitTags queries with FuzzyFilters

ronanh:
This patch reworks how FuzzyFilters are used in the case of ExplicitTags queries:

* Use a list of FuzzyFilters on tag key and value instead of a single FuzzyFilter over tag keys
* No more need for a Rowkey Regex filter (All filtering done by FuzzyFilters + scanner start/stop keys)

It brings noticeable performance improvement when filtering on tag value for metrics with high cardinality (eg. a query for metric for a single host, with host tag cardinality of 20000 which took 9s is down to 250ms)

The list of fuzzy filters is constructed by a new method `QueryUtil.buildFuzzyFilters(row_key_literals)`.
It is used for filtering all combinations of tags provided by the caller.

So for 3 tags `cluster`, `service` and `vm`, with filtering values: `cluster=dev|qa|test`, `service=mysvc` and `vm=web|db`, a list of 8 (= 3 x 1 x 2) FuzzyFilters  covering all combinations will be returned.

A few remarks:

* New (private) version of `QueryUtil.getRowKeyUIDRegex(row_key_literals, explicit_tags)`: simplified variant stripped of everything related to FuzzyFilters
* `QueryUtil.getRowKeyUIDRegex(group_bys, row_key_literals)` and `QueryUtil.getRowKeyUIDRegex(group_bys, row_key_literals, explicit_tags, fuzzy_key, fuzzy_mask)` are left unchanged as these are public API (but not used anymore)

### DIFF
--- a/src/query/QueryUtil.java
+++ b/src/query/QueryUtil.java
@@ -15,6 +15,7 @@ package net.opentsdb.query;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map.Entry;
@@ -26,13 +27,12 @@ import net.opentsdb.core.TSDB;
 import net.opentsdb.uid.UniqueId;
 
 import org.hbase.async.Bytes;
-import org.hbase.async.FilterList;
 import org.hbase.async.FuzzyRowFilter;
+import org.hbase.async.FuzzyRowFilter.FuzzyFilterPair;
 import org.hbase.async.KeyRegexpFilter;
 import org.hbase.async.Bytes.ByteMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.hbase.async.ScanFilter;
 import org.hbase.async.Scanner;
 
 /**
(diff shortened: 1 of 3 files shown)


ronanh:
I also have an [experiment_explicit_tags_gui](https://github.com/ronanh/opentsdb/tree/experiment_explicit_tags_gui) branch in my repo for testing explicit tags from the GUI.

avermeer:
Hi @ronanh, your contrib sounds great to get best performances from OpenTSDB !
Hi @manolama, do you think this PR could be reviewed & integrated if no objection?

Kind regards,
Alexandre

manolama:
Hi @ronanh, what versions of HBase did you test this with? An issue I ran to with older versions was that it failed to actually match on the values, only keys. Thanks!

ronanh:
Yes, I saw a comment in the code referring to the issue you're talking about, but it was not really clear to me what the problem really was. I did not encounter anything like this.
I'm working with HBase from Hortonworks distibution (2.6.5), which is HBase 1.1.2 plus a number of patches. See the details [here](https://docs.cloudera.com/HDPDocuments/HDPforCloud/HDPforCloud-2.6.5/hdp-release-notes/content/patch_hbase.html).

manolama:
Thanks! I think we were still on a 0.98 branch where in setting the filter on the key values caused nothing to return. I'll go through this and make sure it will still handle the case where a user may have a regex or wildcard on a value.

ronanh:
There's been several issues with FuzzyFilters in earlier HBase versions. In my experience it's also quite sensitive to how FuzzyFilter are serialized in the client request, eg. if you use different jar version in client and server (don't know how this applies to AsyncHBase client though)

PR #1907, opened 2020-02-03
3x tweaks

manolama:


common/src/main/java/net/opentsdb/data/ZonedNanoTimeStamp.java
```diff
@@ -76,53 +76,69 @@
...
+    epoch = Instant.from(timestamp).getEpochSecond();
+    nanos = Instant.from(timestamp).getNano();
   }
   
   public ZonedNanoTimeStamp(long epoch_millis, ZoneId zone) {
     if (zone == null) {
       throw new IllegalArgumentException("Zone cannot be null.");
     }
     timestamp = ZonedDateTime.ofInstant(Instant.ofEpochMilli(epoch_millis), zone);
+    epoch = Instant.from(timestamp).getEpochSecond();
```
smrutilal2: You can also do `millisec/1000` to minimize garbage. `Instant.from(timestamp)` would create a new instance of `Instant` every time. In fact twice for the current code. One for second and one more for nano.

manolama: Good catch. I still need to use the timesamp functions though as we need to convert from a zoned timestamp to the Unix epochs.

arungupta:
what are the use cases where you need to use zoned timestamps @manolama?

manolama:
Whenever we have to downsample based on a time zone, e.g. accounting for DST when downsampling over days or weeks.

common/src/main/java/net/opentsdb/pools/ArrayObjectPool.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Attempts to claim an object from the pool immediately, allocating a new
+   * one if they're all in use. If the requested length is greater than the 
+   * configured pool array length, a new array is allocated and will be GC'd.
+   * 
+   * @param length The length of the array to return.
+   * @return A non-null pooled object.
+   */
+  public PooledObject claim(final int length);
```
sidhhu: where is release happening?

manolama: happens in the code that calls the `claim` method.

core/src/main/java/net/opentsdb/pools/BlockingQueueArrayObjectPoolFactory.java
```diff
@@ -0,0 +1,42 @@
...
+ * Generates blocking queue object pools for arrays.
+ * 
+ * @since 3.0
+ */
+public class BlockingQueueArrayObjectPoolFactory extends BaseTSDBPlugin 
+  implements ArrayObjectPoolFactory {
+
+  @Override
+  public ObjectPool newPool(final ObjectPoolConfig config) {
+    return new BlockingQueueObjectPool(tsdb, config);
```
smrutilal2: shouldn't this be `BlockingQueueArrayObjectPool ` instead

core/src/main/java/net/opentsdb/pools/BlockingQueueArrayObjectPool.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+  }
+  
+  @Override
+  public PooledObject claim() {
+    PooledObject obj = pool.poll();
+    if (obj == null) {
+      tsdb.getStatsCollector().incrementCounter("objectpool.claim.miss", 
+          "pool", config.id());
+      return new LocalPooled(config.allocator().allocate(), false);
```
smrutilal2: This seems to be creating an array that will be garbage collected.

manolama: Right, if all of the objects in the pool are checked out then we fall back to allocation.

core/src/main/java/net/opentsdb/pools/BlockingQueueArrayObjectPoolFactory.java
manolama: yup good catch.

core/src/main/java/net/opentsdb/pools/BlockingQueueArrayObjectPool.java
```diff
@@ -0,0 +1,173 @@
...
+    @Override
+    public Object object() {
+      return obj;
+    }
+
+    @Override
+    public void release() {
+      if (was_pooled) {
+        if (!pool.offer(this)) {
+          LOG.warn("Failed to return a pooled object to the pool for " + config.id());
```
smrutilal2: should decrement the count

manolama: oh and I don't actually have a count here to decrement. The config does have it but that's just an initialization time value. I can add a "size" getter if we need it.

common/src/main/java/net/opentsdb/pools/ArrayObjectPoolFactory.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface ArrayObjectPoolFactory extends TSDBPlugin {
+
+  /**
+   * Instantiates a new pool with the given config.
+   * @param config A non-null config.
+   * @return A non-null pool or an exception if the config was missing required
+   * information.
+   */
+  public ObjectPool newPool(final ObjectPoolConfig config);
```
smrutilal2: return type should be `ArrayObjectPool `

PR #1914, opened 2020-02-11
COMMON:

manolama:
- Truly fix up the PluginLoader for Docker and JDK11. Uses the Instrumentation
  method as described by Chris Jennings:
  https://cgjennings.ca/articles/java-9-dynamic-jar-loading/

### DIFF
--- a/common/src/main/java/net/opentsdb/utils/PluginLoader.java
+++ b/common/src/main/java/net/opentsdb/utils/PluginLoader.java
@@ -17,6 +17,7 @@ package net.opentsdb.utils;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.lang.instrument.Instrumentation;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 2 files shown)


common/src/main/java/net/opentsdb/utils/PluginLoader.java
```diff
@@ -395,15 +410,43 @@ private static void addFile(File f) throws IOException, SecurityException,
...
+        INSTRUMENTATION.appendToSystemClassLoaderSearch(new JarFile(url.getFile()));
+        LOG.debug("Successfully added JAR to class loader: " + url.getFile());
+      } catch (IOException e) {
+        throw new IllegalArgumentException("Failed to load JAR: " + url, e);
+      }
+      return;
+    } else if (ClassLoader.getSystemClassLoader() instanceof URLClassLoader) {
+      // Yay! The old pre-JDK9 hack will most likely work      
+      class_loader = ClassLoader.getSystemClassLoader();
+    } else if (Thread.currentThread().getContextClassLoader() instanceof URLClassLoader) {
```
SeanPMiller: Ideally, these last two branches would just be merged into a big error message and an `exit()` call, but for test-case purposes, I'm fine with having them for now.

PR #1915, opened 2020-02-11
CORE:

manolama:
- Fix another array index bug in GroupByNumericArrayIterator.


core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -297,7 +297,7 @@ private void accumulateInParallel(
 
     final CountDownLatch doneSignal = new CountDownLatch(jobCount);
     for (int jobIndex = 0; jobIndex < jobCount; jobIndex++) {
-      NumericArrayAggregator combiner = combiners[jobIndex % jobCount];
+      NumericArrayAggregator combiner = combiners[jobIndex % combiners.length];
```
smrutilal2: we need this fix in `GroupByNumericSummaryParallelIterator` as well.

manolama: fixed

PR #1919, opened 2020-02-21
ELASTICSEARCH:

manolama:
-Set the data result to no-data if we filter out all of the results.


implementation/elasticsearch/src/main/java/net/opentsdb/meta/impl/es/ESMetaResponse.java
```diff
@@ -737,6 +738,10 @@ NamespacedAggregatedDocumentResult parseTimeseries(
         result.addTimeSeries(id, meta_query, metric, false);
       }
     }
+    
+    if (isMultiGet && result.timeSeries().isEmpty()) {
```
mailravi3390: Can the result be null at this point ?

manolama: Nope, safe here as there's a null check in the getter.

PR #1920, opened 2020-02-25
3x tweaks

manolama:


core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAverageFactory.java
```diff
@@ -31,73 +34,98 @@
...
-
+    
     protected int[] counts;
+    protected PooledObject int_pooled;
     protected double[] results;
     
-    public ArrayAverage(final boolean infectious_nans) {
-      super(infectious_nans);
+    public ArrayAverage(final boolean infectious_nans,
+                        final BaseArrayFactory factory) {
```
smrutilal2: type of the 2nd argument can be `ArrayAverageFactory` instead.

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAverageFactory.java
```diff
@@ -31,73 +34,98 @@
...
+      super(infectious_nans, factory);
+    }
+    
+    public ArrayAverage(final NumericArrayAggregatorConfig config,
+                        final BaseArrayFactory factory) {
```
smrutilal2: `ArrayAverageFactory `

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAverageFactory.java
```diff
@@ -31,73 +34,98 @@
...
+        initLong(values, from, to);
+        
+        if (((ArrayAverageFactory) factory).intPool() != null) {
+          int_pooled = ((ArrayAverageFactory) factory).intPool().claim(to - from);
+          counts = (int[]) int_pooled.object();
+        } else {
+          counts = new int[to - from];
+        }
+        
+        Arrays.fill(counts, 0, end, 1);
```
smrutilal2: why are we filling the array with `1`?

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAverageFactory.java
```diff
@@ -229,7 +348,17 @@ public boolean isInteger() {
...
     public double[] doubleArray() {
       if (results == null) {
-        results = new double[counts.length];
+        PooledObject double_pooled = null;
+        if (double_accumulator != null) {
+          // reuse when we can.
+          results = double_accumulator;
+        } else if (factory.doublePool() != null) {
+          double_pooled = factory.doublePool().claim(end);
+          results = (double[]) double_pooled.object();
```
smrutilal2: don't you have to reset the pooled array like other places above.

core/src/main/java/net/opentsdb/query/interpolation/types/numeric/NumericInterpolator.java
```diff
@@ -176,6 +176,17 @@ public TimeStamp nextReal() {
     return fill_policy;
   }
 
+  @Override
+  public void close() {
+    try {
+      iterator.close();
+    } catch (IOException e) {
+      // Don't bother logging.
+      e.printStackTrace();
```
smrutilal2: may be remove the `printStackTrace` too.

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAverageFactory.java
manolama: When it's a long array that means we have a value for every count.

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAverageFactory.java
manolama: No because we'll be setting the value for every entry in the array on line 367 so we don't need to reset here.

core/src/main/java/net/opentsdb/query/interpolation/types/numeric/NumericInterpolator.java
manolama: I do want to keep these around just to see if they're every executed. They shouldn't be but it's possible at some point in the future.

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAverageFactory.java
manolama: Hmm I could have but I'd rather not go back and rebase this whole commit.

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAverageFactory.java
manolama: Hmm I could have but I'd rather not go back and rebase this whole commit.

PR #1963, opened 2020-06-03
CORE:

manolama:
- Copy over the query planner dat a sources code from the DefaultQueryPlanner
  into the read cache context and tweak it for proper countdowns.


core/src/main/java/net/opentsdb/query/ReadCacheQueryPipelineContext.java
```diff
@@ -1329,6 +1347,81 @@ public int compareTo(Object o) {
...
+      return Lists.newArrayList(((MergerConfig) node).getDataSource() + ":" 
+          + ((MergerConfig) node).getDataSource());
+    } else if (node.joins()) {
+      return Lists.newArrayList(node.getId() + ":" + node.getId());
+    } else if (node instanceof TimeSeriesDataSourceConfig) {
+      return Lists.newArrayList(node.getId() + ":" 
+          + ((TimeSeriesDataSourceConfig) node).getDataSourceId());
+    }
+    
+    List<String> sources = null;
```
smrutilal2, line 1366: the variable `sources` is only used inside the `if` block at line number: 1369. Move it inside the `if` block please

PR #1663, opened 2019-06-04
Make UniqueIdRpc aware of the mode

eswdd:
First stab at implementing following comments on https://groups.google.com/forum/#!topic/opentsdb/NDRNuDLesFY

PR #1966, opened 2020-06-12
Pr 1663

johann8384:


johann8384:
This replaces #1663, basing off next and testing.

PR #1977, opened 2020-07-02
COMMON:

manolama:
- QueryInterpolator now implements a TypedTimeSeriesIterator

CORE:
- Fix up JsonV2ExpQuerySerdes with proper synchronization (TODO - improve),
  millisecond timestamps like in V2 and support for the NumericArrayType. More
  to do still.
- Fix a couple of bugs in the NumericType JsonReadCacheSerdes path.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/execution/serdes/JsonV2ExpQuerySerdes.java
+++ b/core/src/main/java/net/opentsdb/query/execution/serdes/JsonV2ExpQuerySerdes.java
@@ -170,7 +170,6 @@ public class JsonV2ExpQuerySerdes implements TimeSeriesSerdes {
       @Override
       public synchronized Object call(final ArrayList<TimeSeriesStringId> ids) 
             throws Exception {
-        
         QueryInterpolatorFactory factory = tsdb
             .getRegistry().getPlugin(QueryInterpolatorFactory.class, 
                 NIC.getType());
@@ -319,7 +318,6 @@ public class JsonV2ExpQuerySerdes implements TimeSeriesSerdes {
               "Unexpected exception "
               + "serializing: " + result, 500, e));
         }
-        
         return Deferred.fromResult(null);
       }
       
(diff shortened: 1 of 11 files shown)


core/src/main/java/net/opentsdb/query/execution/serdes/JsonV2ExpQuerySerdes.java
```diff
@@ -389,6 +349,147 @@ public Object call(final Exception ex) throws Exception {
...
+          json.writeNumber(value.longArray()[value.offset() + index]);
+        } else {
+          json.writeNumber(value.doubleArray()[value.offset() + index]);
+        }
+      }
+    
+      index++;
+      next_ts.add(result.timeSpecification().interval());
+      json.writeEndArray();
+      if (next_ts.compare(Op.GTE, context.query().endTime())) {
```
gskchaitanya, line 487: If the query endtime is in the future, the next_ts will never be greater than query endtime and will it cause an infi loop?

manolama: Ah good catch, I missed the has_next call in here.

manolama: Actually the other catch is on L473 where it will bail when we reach the end of the array.

core/src/main/java/net/opentsdb/query/pojo/TimeSeriesQuery.java
```diff
@@ -730,6 +732,34 @@ public boolean hasKey(final String key) {
...
+    // Set the outputs properly
+    if (outputs != null && !outputs.isEmpty()) {
+      final List<String> outs = Lists.newArrayListWithExpectedSize(
+          outputs.size());
+      // TODO - alias
+      for (final Output output : outputs) {
+        outs.add(output.getId());
+      }
+      builder.addSerdesConfig(JsonV2QuerySerdesOptions.newBuilder()
+          .setId("JsonV2ExpQuerySerdes")
```
sidhhu, line 745: want to pull these out as static strings?

PR #1990, opened 2020-07-29
3x tweaks

manolama:


storage/asynchbase/src/main/java/net/opentsdb/storage/Tsdb1xMultiGet.java
```diff
@@ -448,13 +469,13 @@ void onError(final Throwable t) {
...
       state = State.EXCEPTION;
-        if (!node.push()) {
+      if (!node.push()) {
         current_result.setException(t);
         final QueryResult result = current_result;
         current_result = null;
         outstanding.set(0);
         node.onNext(result);
-      } else {
+      } else if (!node.pipelineContext().queryContext().isClosed()) {
```
sidhhu, line 478: call error if query context is closed?

implementation/servlet/src/main/java/net/opentsdb/servlet/resources/AsyncRunner.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Static helper to handle an async query. 
+ * 
+ * @since 3.0
+ */
+public class AsyncRunner {
+  private static final Logger LOG = LoggerFactory.getLogger(AsyncRunner.class);
+  
+  private AsyncRunner() { }
```
sidhhu, line 42: Do you need this?

storage/asynchbase/src/main/java/net/opentsdb/storage/Tsdb1xMultiGet.java
manolama: Nope, cause we no longer have any place to send the error.

implementation/servlet/src/main/java/net/opentsdb/servlet/resources/AsyncRunner.java
manolama: Yeah cause I don't want anyone instantiating it. It's how you mark a static class as intending to hold static methods only..

PR #1996, opened 2020-08-20
3x tweaks

manolama:


core/src/test/java/net/opentsdb/query/processor/groupby/TestGroupByNumericSummaryParallelIterator.java
```diff
@@ -92,12 +92,20 @@ public static void beforeClass() {
...
     when(FACTORY.predicate()).thenReturn(p);
     when(FACTORY.tsdb()).thenReturn(TSDB);
+    TSDB.getRegistry().registerFactory(FACTORY);
     
-    GroupByJobPool allocator = FACTORY.new GroupByJobPool();
+    GroupByJobPool allocator = new GroupByJobPool();
+    try {
+      allocator.initialize(TSDB, null).join();
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
smrutilal2, line 101: log error instead

core/src/main/java/net/opentsdb/pools/BlockingQueueArrayObjectPool.java
```diff
@@ -103,8 +103,10 @@ public PooledObject claim(final long time, final ChronoUnit unit) {
   @Override
   public PooledObject claim(final int length) {
     if (length > ((ArrayObjectPoolAllocator) config.allocator()).pooledLength()) {
-      tsdb.getStatsCollector().incrementCounter("objectpool.claim.largeArray", 
+      tsdb.getStatsCollector().incrementCounter("objectpool.claim.lengthTooBig.count", 
           "pool", config.id());
+      tsdb.getStatsCollector().setGauge("objectpool.claim.lengthTooBig.size",
```
smrutilal2, line 108: This is the metric to track array size.  👍

core/src/test/java/net/opentsdb/query/processor/groupby/TestGroupByNumericSummaryParallelIterator.java
manolama: naw just a UT.

core/src/main/java/net/opentsdb/pools/BlockingQueueArrayObjectPool.java
smrutilal2: The `Gauge` metric will give the count as well. We don't need another `Counter` metric

PR #1997, opened 2020-08-22
COMMON:

manolama:
- Add the QueryResultId and a default implementation. This allows for much cleaner
  and simpler result tracking through the query execution graph. It replaces the
  DataSourceId fields and lets the nodes set their own results so it is super
  easy for nodes (like the expression nodes) to find out what results are
  expected. Also add the appropriate builder methods and config getters.
- Also add a getter/setter, markedCacheable and markCacheable that will be used
  to replace the current read cache setup.

CORE:
- Implement the new QueryResultId everywhere.
- Change the DefaultQueryPlanner config node setup to properly walk the graph
  depth first so the upstream nodes have all of their downstream nodes setup.
  Also split up the plan() method into separate bits we can use in the read
  cache.
- Add a bit of logging to the Joiner.
- Fix up the toBuilder() calls for nodes that were missing implementations and
  add some helpers to the base classes.

### DIFF
--- a/common/src/main/java/net/opentsdb/query/DefaultQueryResultId.java
+++ b/common/src/main/java/net/opentsdb/query/DefaultQueryResultId.java
@@ -1,0 +1,87 @@
+// This file is part of OpenTSDB.
+// Copyright (C) 2020  The OpenTSDB Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package net.opentsdb.query;
+
+import com.google.common.base.Strings;
+
+/**
+ * A default implementation of the query result ID.
+ * 
+ * @since 3.0
+ */
+public class DefaultQueryResultId implements QueryResultId {
+  /** The node ID, part 1 of the colon delimited ID string. */
+  private final String node_id;
+  
+  /** The data source for the result, part 2 of the colon delimited ID String. */
+  private final String data_source;
+  
+  /**
+   * Default ctor.
+   * @param node_id The non-null and non-empty node ID.
+   * @param data_source The non-null and non-empty data source.
+   */
+  public DefaultQueryResultId(final String node_id, final String data_source) {
+    if (Strings.isNullOrEmpty(node_id)) {
+      throw new IllegalStateException("Node ID cannot be null or empty.");
+    }
+    if (Strings.isNullOrEmpty(data_source)) {
+      throw new IllegalStateException("Data source cannot be null or empty.");
+    }
+    this.node_id = node_id;
+    this.data_source = data_source;
+  }
+  
+  @Override
+  public String nodeID() {
+    return node_id;
+  }
+
+  @Override
+  public String dataSource() {
+    return data_source;
+  }
+
+  @Override
+  public String toString() {
+    return node_id + ":" + data_source;
+  }
+  
+  @Override
+  public boolean equals(final Object obj) {
+    if (obj == null || !(obj instanceof QueryResultId)) {
+      return false;
+    }
+    
+    final QueryResultId other = (QueryResultId) obj;
+    if (Strings.isNullOrEmpty(other.nodeID()) || 
+        Strings.isNullOrEmpty(other.dataSource())) {
+      throw new IllegalStateException("ID " + obj 
+          + " has a null or empty node or data source ID.");
+    }
+    
+    return other.nodeID().equals(node_id) &&
+           other.dataSource().equals(data_source);
+  }
+  
+  @Override
+  public int hashCode() {
+    // TODO - verify this is ok for ints. It should be as we'd have low
+    // cardinality within a given query. If we ever use this for multi-query
+    // then we need to revisit it.
+    return 2251 * node_id.hashCode() ^ 37 * data_source.hashCode();
+  }
+  
+}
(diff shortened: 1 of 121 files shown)


common/src/main/java/net/opentsdb/query/DefaultQueryResultId.java
```diff
@@ -0,0 +1,87 @@
...
+  @Override
+  public boolean equals(final Object obj) {
+    if (obj == null || !(obj instanceof QueryResultId)) {
+      return false;
+    }
+    
+    final QueryResultId other = (QueryResultId) obj;
+    if (Strings.isNullOrEmpty(other.nodeID()) || 
+        Strings.isNullOrEmpty(other.dataSource())) {
+      throw new IllegalStateException("ID " + obj
```
smrutilal2, line 71: shouldn't this return false instead of throwing exception.

core/src/main/java/net/opentsdb/pools/ByteArrayPool.java
```diff
@@ -50,6 +50,7 @@ public String type() {
         .setAllocator(this)
         .setInitialCount(tsdb.getConfig().getInt(configKey(COUNT_KEY, TYPE)))
         .setMaxCount(tsdb.getConfig().getInt(configKey(COUNT_KEY, TYPE)))
+        .setArrayLength(tsdb.getConfig().getInt(configKey(LENGTH_KEY, TYPE)))
```
smrutilal2, line 53: Is this fixing the default length of byte array to 8K

core/src/main/java/net/opentsdb/pools/ByteArrayPool.java
```diff
@@ -77,10 +78,10 @@ public Object allocate(int length) {
   
   @Override
   protected void registerConfigs(final Configuration config, final String type) {
-    super.registerConfigs(config, type);
     if (!config.hasProperty(configKey(LENGTH_KEY, TYPE))) {
       config.register(configKey(LENGTH_KEY, TYPE), 8192, false, 
           "The length of each array to allocate");
     }
+    super.registerConfigs(config, type);
```
smrutilal2, line 85: I see. So, this was the issue.

common/src/main/java/net/opentsdb/query/readcache/CachedQueryNode.java
```diff
@@ -136,6 +137,22 @@ public Builder toBuilder() {
     return null;
   }
 
+  @Override
+  public List<QueryResultId> resultIds() {
+    // TODO - do we need to populate this?
+    return Collections.emptyList();
```
smrutilal2, line 143: why not return a null. I understand, the client will have to do a null check. But, it will avoid the object creation.

core/src/main/java/net/opentsdb/query/BaseQueryNodeConfig.java
```diff
@@ -244,6 +251,35 @@ public HashCode buildHashCode() {
     return hc.hash();
   }
   
+  @Override
+  public List<QueryResultId> resultIds() {
+    return result_ids == null ? Collections.emptyList() : result_ids;
```
smrutilal2, line 256: As commented before, Let's just return `null`

core/src/main/java/net/opentsdb/query/ReadCacheQueryPipelineContext.java
```diff
@@ -337,17 +339,24 @@ public Void call(ArrayList<Void> arg) throws Exception {
...
-          if (serdes_filter != null && !serdes_filter.isEmpty()) {
-            if (serdes_filter.contains(source)) {
-              countdowns.put(source, new AtomicInteger(sinks.size()));  
-            }
-          } else {
-            countdowns.put(source, new AtomicInteger(sinks.size()));
-          }
+// TODO --------------------------*(*        
+        CacheQueryPlanner planner = new CacheQueryPlanner(
+            ReadCacheQueryPipelineContext.this, (QueryNode) ReadCacheQueryPipelineContext.this);
```
smrutilal2, line 344: Why are you passing the same object twice as the argument?

core/src/main/java/net/opentsdb/query/ReadCacheQueryPipelineContext.java
```diff
@@ -337,17 +339,24 @@ public Void call(ArrayList<Void> arg) throws Exception {
...
-        for (final String source : serdes_sources) {
-          if (serdes_filter != null && !serdes_filter.isEmpty()) {
-            if (serdes_filter.contains(source)) {
-              countdowns.put(source, new AtomicInteger(sinks.size()));  
-            }
-          } else {
-            countdowns.put(source, new AtomicInteger(sinks.size()));
-          }
+// TODO --------------------------*(*        
+        CacheQueryPlanner planner = new CacheQueryPlanner(
```
smrutilal2, line 343: Is it going to create one `CacheQueryPlanner` object for each query? Can this be reused?

common/src/main/java/net/opentsdb/query/DefaultQueryResultId.java
manolama: No cause it's a coding issue at this stage.

core/src/main/java/net/opentsdb/pools/ByteArrayPool.java
manolama: yup

common/src/main/java/net/opentsdb/query/readcache/CachedQueryNode.java
manolama: It uses the static empty list here.

core/src/main/java/net/opentsdb/query/BaseQueryNodeConfig.java
manolama: I tried the null route and it really is ugly. The static ref is a bit cleaner.

core/src/main/java/net/opentsdb/query/ReadCacheQueryPipelineContext.java
manolama: In this case because the parent implements both. In some cases it may not.

core/src/main/java/net/opentsdb/query/ReadCacheQueryPipelineContext.java
manolama: Yeah for each query right now. Eventually we can pool it if it becomes a problem.

common/src/main/java/net/opentsdb/query/readcache/CachedQueryNode.java
smrutilal2: You are right. That should be fine then.

core/src/main/java/net/opentsdb/query/BaseQueryNodeConfig.java
smrutilal2: Agreed.

PR #2001, opened 2020-08-28
CORE:

manolama:
- Fix an issue with summaries in the read cache pipeline
- Fix a CME in the query planner and add some result IDs to the debug log
- Yank the override for the Merger config result IDs as it actually broke
  the naming.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/plan/DefaultQueryPlanner.java
+++ b/core/src/main/java/net/opentsdb/query/plan/DefaultQueryPlanner.java
@@ -971,7 +971,8 @@ public class DefaultQueryPlanner implements QueryPlanner {
       }
       
       final List<QueryNodeConfig> push_downs = Lists.newArrayList();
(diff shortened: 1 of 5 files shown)


core/src/main/java/net/opentsdb/query/plan/DefaultQueryPlanner.java
```diff
@@ -971,7 +971,8 @@ protected void setupPushDowns() {
       }
       
       final List<QueryNodeConfig> push_downs = Lists.newArrayList();
-      for (final QueryNodeConfig n : config_graph.predecessors(node)) {
+      final Set<QueryNodeConfig> nodes = Sets.newHashSet(config_graph.predecessors(node));
```
smrutilal2, line 974: why not iterate on the `config_graph.predecessors(node)` itself

manolama: yeah, explicit dedupe to avoid a concurrent modification exception as the call below can mutate the graph.

PR #2011, opened 2020-09-07
ES:

manolama:
- Add some more NPE catches in the response. Something is up with this
  ancient library.


implementation/elasticsearch/src/main/java/net/opentsdb/meta/impl/es/ESMetaResponse.java
```diff
@@ -96,7 +96,27 @@ public ESMetaResponse(final Map<String, MultiSearchResponse> response) {
...
+          LOGGER.error("Null or empty responses for response: " + response_entry.getKey());
+          continue;
+        }
+        
         final SearchResponse response = response_entry.getValue().getResponses()[0].getResponse();
+        if (response == null) {
+          LOGGER.error("Null SearchResponse at index 0 for response: " + response_entry.getKey());
+          continue;
+        }
+        if (response.getHits() == null) {
```
sidhhu, line 115: Is this even possible?

PR #2012, opened 2020-09-08
ES:

sidhhu:
Implement a basic search feature in Meta we can search for a string in either tag keys or values and display the matching key-values


manolama (approved):
OK we'll try it out, maybe tweak it a bit in the future.

PR #2014, opened 2020-09-09
3x tweaks

manolama:


core/src/main/java/net/opentsdb/utils/XXHash.java
```diff
@@ -0,0 +1,65 @@
...
+  public static long hash(final byte[] value, final int offset, final int length) {
+    return HASHER.hashBytes(value, offset, length);
+  }
+  
+  public static long hash(final String value) {
+    return HASHER.hashChars(value);
+  }
+  
+  public static long updateHash(final long hash, final byte[] value) {
+    return 2251 * hash ^ 37 * hash(value);
```
smrutilal2, line 47: `combineHashes(hash, hash(value))`

core/src/main/java/net/opentsdb/utils/XXHash.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+  
+  public static long updateHash(final long hash, 
+                                final byte[] value, 
+                                final int offset, 
+                                final int length) {
+    return 2251 * hash ^ 37 * hash(value, offset, length);
```
smrutilal2, line 54: combineHashes(hash, hash(value, offset, length))

core/src/main/java/net/opentsdb/utils/XXHash.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+  
+  public static long updateHash(final long hash, final String value) {
+    return 2251 * hash ^ 37 * hash(value);
```
smrutilal2, line 58: `combineHashes(hash, hash(value))`

core/src/main/java/net/opentsdb/utils/XXHash.java
manolama: Actually I'd like to avoid the redirect so we don't have to wait for 1,000 calls to inline the method. I'll also yank the hash redirect as well.

smrutilal2 (approved):

Issue #1635, opened 2019-05-13
OpenTSDB 2.4.0 query limits ineffective. High running queries alerts


HariSekhon:
We've been periodically suffering alerts of high count of running queries on our production cluster, with a restart of affected nodes being the short term workaround to reset it, but it can recur any time.

We believe because of large queries sent from Grafana which the whole company uses heavily (we ingest billions of data points a day and run millions of queries a day on it). Running queries are usually in the 1-30 range on each of our TSDB instances behind our load balancer, but for example this morning 3 nodes had unusually high counts of 32200, 29400 and 9500 running queries on each of those 3 TSDS. The other TSDs were fine with only a handful of running queries (the LB is set to round robin). In cases where some TSDs end up in this state it can cause lag on our dashboards, and the graph shows the numbers gradually increases over time like they are accumulating a backlog of queries stuck behind some bigger queries in the HBase call queue, but this is hard to prove with the current lack of OpenTSDB instrumentation about currently running queries.

It's very difficult to track down what big running queries are backing up the call queues in HBase which is why I raised a ticket for feature request for a running queries list #1624.

So we upgraded from OpenTSDB 2.3 to OpenTSDB 2.4 last week and added the following query limits in the hope of preventing this recurring issue:

```
tsd.query.timeout = 1000
tsd.query.limits.bytes.default = 3048576000
tsd.query.limits.data_points.default = 1000000
```

We had originally set lower bytes default eg. 350MB but this blocked too many historical queries and eventually upped it to 3GB which is a relatively small drop in the ocean of our cluster which is dozens of TB.

I suspect these limits aren't being enforced properly, at least for the query timeout, and others have noticed the same as dashboard graphs do sometimes taken noticeably longer than 1 second without getting killed, although not very often because I tuned our HBase cluster to the hilt a few months ago.

HariSekhon:
Regarding finding in flight queries, the logs shows new queries and completed queries as well as

```
2019-05-13T09:10:40.407+0000 INFO  [HttpQuery.logInfo] - [id: 0xda2458d9, /<ip_x.x.x.x> => /<ip_x.x.x.x>:4242] HTTP /api/query done in 4ms
```

but the hex ID in the `done in X ms` line isn't shown in the:

```
2019-05-13T09:10:40.414+0000 INFO  [QueryStats.<init>] - Executing new query={"query": ...
```

or

```
[...]

PR #2035, opened 2020-10-26
Re-introduce query timeouts.

johann8384:
Fixes #1830


itamarst:
Thank you! I forget the details, but does this also fix https://github.com/OpenTSDB/opentsdb/issues/1635?

johann8384:
I will have to see, I am not familiar with that one.

Issue #2056, opened 2020-12-04
Uri should not transfer to lower case in FileFactory.java


pengmq1:
when start 3.0 tsd use:
./bin/tsdb tsd --config.providers=file:///Var/opentsdb.yaml
it will throw ERROR and terminate.
WARN  [main] YamlJsonFileProvider: No file found at: /var/opentsdb.yaml
but the is truely exist in /Var/opentsdb.yaml. 
Because the file PATH has Upper case character.
It is a bug.

pengmq1:
see PR as follows

manolama:
Merged, thanks.

PR #2057, opened 2020-12-04
Fix bug: uri should not transfer to lower case. (#2056)

pengmq1:


manolama:
Good catch, thank you.

PR #2064, opened 2020-12-15
CORE:

manolama:
- Fix the ratio node so it can find the proper data source.


SeanPMiller (approved):
Test case in 2nd commit looks good.

PR #2073, opened 2021-01-15
3x tweaks

manolama:


core/src/main/java/net/opentsdb/query/processor/expressions/ExpressionNumericArrayIterator.java
```diff
@@ -41,9 +41,33 @@
     BaseExpressionNumericIterator<NumericArrayType>
   implements NumericArrayType {
 
+  private static final int STATIC_ARRAY_LEN = 86_400;
```
smrutilal2, line 44: Why and array length of `86_400 `. Isn't that to much of allocation for small time range queries?

manolama: Just wanted to capture most cases. I'll add some instrumentation eventually and see what the best size is. 1 day of second data should be adequate though :)

smrutilal2: And we have ThreadLocal allocation. Which is 8 by default, I guess.

smrutilal2: I guess, it depends how big is the actual data segment. If the segment is smaller For example 2 hours incase of Aura, then `7200` is probably good enough.

smrutilal2 (approved):

PR #2093, opened 2021-03-22
Add support to read configurations from an InputStream

### LINKED ISSUE #2088: Read configurations from InputStream
Currently, `net.opentsdb.utils.Config` has a constructor that accepts a path to the OpenTSDB configuration file.
```java
/**
 * Constructor that initializes default values and attempts to load the given
 * properties file
 * @param file Path to the file to load
 * @throws IOException Thrown if unable to read or parse the file
 */
public Config(final String file) throws IOException {
  loadConfig(file);
  setDefaults();
}
```

Since the structure of an application changes when it is packaged as a JAR file, this path would no longer be valid for applications using this constructor that are deployed as JAR files. (assuming the configuration file is store under src/main/resources)

Would it make sense to add another constructor that accepts an `InputStream` of the contents of the configuration file instead of just a path to the file? Or are there any other ways to set the configurations in a JAR if the file is placed under src/main/resources?

I'd be happy to open a PR for the same if this sounds like a valid usecase.

melburne:
Closes #2088

### DIFF
--- a/common/src/main/java/net/opentsdb/utils/Config.java
+++ b/common/src/main/java/net/opentsdb/utils/Config.java
@@ -17,6 +17,7 @@ package net.opentsdb.utils;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


melburne:
Hi @manolama 
I've added the `public Config(InputStream input_stream)` constructor to read configurations from an InputStream. 
Would be great to get your feedback on this.

manolama (approved):

PR #2107, opened 2021-04-19
CORE:

manolama:
- Fix the by-ref in GBNAI and wrap some more try catch statements.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
+++ b/core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalAmount;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
```diff
@@ -347,7 +348,7 @@ public void run() {
             // AGG
             try {
               final TimeSeries source = sources != null ? sources.get(startIndex) :
-                result.downstreamResult().timeSeries().get(startIndex);
+                result.downstreamResult().timeSeries().get(arraySources[startIndex]);
```
mailravi3390, line 351: Is it possible for arraySources to be null ? (I see one of the constructors setting it to null).

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByTimeSeries.java
```diff
@@ -141,7 +141,7 @@ void addSource(final TimeSeries source, final int idx) {
     }
     // TODO - pool
     if (source_refs == null) {
-      source_refs = new int[16];
+      source_refs = new int[8];
```
smrutilal2, line 144: it starts with initial size of 8 and then keeps growing?

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByNumericArrayIterator.java
manolama: Yeah, that's why we have the `sources != null` check.

core/src/main/java/net/opentsdb/query/processor/groupby/GroupByTimeSeries.java
manolama: Yup, reducing it for the case wherein we have 1 series per group. Still need to pool it eventually.

PR #2111, opened 2021-04-20
ES: Plugin support for TagValueLiteralOr Filter

anonthing:


implementation/elasticsearch/src/main/java/net/opentsdb/meta/NamespacedAggregatedDocumentQueryBuilder.java
```diff
@@ -207,9 +206,12 @@ FilterBuilder getMetricFilter(final MetricFilter filter, final boolean nested) {
     }
   }
 
-  FilterBuilder getTagValueFilter(final TagValueFilter filter, final boolean nested) {
+
+
+  FilterBuilder getTagValueFilter(final TagValueFilter filter, final boolean nested) throws Exception {
     if (filter instanceof TagValueLiteralOrFilter) {
       // handles the range filter as well.
+      filter.initialize(null).join();
```
manolama, line 214: Could you toss the Exceptions in the signatures and just catch and throw a new QueryException please?

manolama (changes requested):
Should be a quick tweak please.

manolama (approved):
LGTM

PR #2116, opened 2021-04-29
3x tweaks

manolama:


implementation/prometheus/src/main/java/net/opentsdb/servlet/resources/PromQLResource.java
```diff
@@ -245,8 +248,9 @@ private Response handleQuery(final ServletConfig servlet_config,
           .build();
     }
     
-    long query_timeout = (long) servlet_config.getServletContext()
+    int async_timeout = (Integer) servlet_config.getServletContext()
```
smrutilal2, line 251: cast it to `int` instead of `Integer`

implementation/ultrabrew/pom.xml
```diff
@@ -13,15 +13,7 @@
...
    <packaging>jar</packaging>
-   
-   <repositories>
-     <repository>
-       <id>bintray-ultrabrew-maven</id>
-       <name>bintray</name>
-       <url>https://dl.bintray.com/ultrabrew/m2</url>
-     </repository>
-   </repositories>
-
```
smrutilal2, line 24: 👍

implementation/prometheus/src/main/java/net/opentsdb/servlet/resources/PromQLResource.java
manolama: It's already auto boxed to an `Integer` due to being in the servlet map so it's ok to keep it like this. The cast from `Integer` to `long` was broken though.

PR #2119, opened 2021-05-06
3.0 Rollup Writes

manolama:


common/src/main/java/net/opentsdb/storage/WriteStatus.java
```diff
@@ -24,6 +24,10 @@
  */
 public interface WriteStatus {
 
+  static final String OK_STRING = "state=OK";
+  static final String RETRY_STRING = "state=OK";
```
smrutilal2, line 28: Is it a typo? should it be `state=RETRY`

smrutilal2 (approved):

PR #2127, opened 2021-05-08
Fix remote code execution #2051 by adding regex validators for the

manolama:
Gnuplot params and introducting the tsd.gnuplot.options.allowlist
setting that is a strict matching allow list of o= values from the
query string that will be allowed through. By default tihs is empty
so if folks are using this query param, they'll different graphs
until they add the options they need.


anonthing (approved):
Good catch!

PR #2128, opened 2021-05-11
COMMON:

manolama:
- Implement common classes for Aura metrics.


common/src/main/java/net/opentsdb/collections/DirectByteArray.java
```diff
@@ -0,0 +1,278 @@
...
+  public void free() {
+    if (address > 0) {
+      unsafe.freeMemory(address);
+      address = 0;
+      capacity = 0;
+    }
+  }
+
+  @Override
+  protected void finalize() {
```
smrutilal2, line 274: I have removed the `finalizer`. Looks like you are moving a older version.

common/src/test/java/net/opentsdb/collections/TestDirectByteArray.java
```diff
@@ -0,0 +1,266 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package net.opentsdb.collections;
+
+import java.util.Random;
+
+public class TestDirectByteArray {
+//  @Test
```
smrutilal2, line 20: why are the tests commented?

manolama: Cause TSD is still using junit 4x. I need to move to 5x

PR #2131, opened 2021-05-18
Starting on split queries.

manolama:


anonthing (approved):

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAggregatorUtils.java
```diff
@@ -0,0 +1,2 @@
+package net.opentsdb.data.types.numeric.aggregators;public class ArrayAggregatorUtils {
```
smrutilal2, line 1: Looks like unintentionally added.

manolama: doh yeah I'll fix that, thanks.

PR #2137, opened 2021-06-10
Adding EnvoyHeaderAuth filter

ntippabhatla:
Please let me know if I have to make any changes


implementation/athenz/src/main/java/net/opentsdb/auth/EnvoyHeaderAuthFilter.java
```diff
@@ -0,0 +1,207 @@
+// This file is part of OpenTSDB.
```
manolama, line 1: You can delete this one, thanks!

implementation/servlet/src/main/java/net/opentsdb/servlet/filter/EnvoyHeaderAuthFilter.java
```diff
@@ -0,0 +1,207 @@
...
+              "Missing or invalid header ");
+    }
+  }
+
+  @Override
+  public AuthState authenticate(final ServletRequest servletRequest) {
+
+    String header=((HttpServletRequest ) servletRequest ).getHeader(tsdb.getConfig().getString(ENVOY_HEADER_KEY));
+    if (header != null){
+      String commonName=header.split(";")[2].split(",")[0].split("=")[2];
```
manolama, line 115: This can throw NPEs if the header is malformed. Split it into multiple steps wherein you split, validate the result is the proper size, then pick the field you want, validate it's name, then split if applicable, repeat till you get the value you want. If a field is missing or invalid then you can log at the debug level and return a null.

manolama (changes requested):
Two tweaks and it should be good, thanks!

implementation/servlet/src/main/java/net/opentsdb/servlet/filter/EnvoyHeaderAuthFilter.java
ntippabhatla: Yes, ack.!

implementation/servlet/src/main/java/net/opentsdb/servlet/filter/EnvoyHeaderAuthFilter.java
```diff
@@ -110,22 +112,39 @@ protected void runFilter(final ServletRequest servletRequest,
...
-        @Override
-        public String getName() {
-          return commonName;
+    String header = ((HttpServletRequest) servletRequest).getHeader(tsdb.getConfig().getString(ENVOY_HEADER_KEY));
+    if (header != null) {
+      if (header.contains("Subject")){
+        String headerSplit=header.substring(header.indexOf("Subject"),header.length());
+        if (headerSplit.contains("CN")){
+          String metaInfo=headerSplit.substring(headerSplit.indexOf("CN"),headerSplit.indexOf(","));
+          String commonName=metaInfo.split("=")[1];
```
manolama, line 121: Still a possible IOBE here but we can try it for now, thanks!

manolama (approved):

PR #2138, opened 2021-06-14
Is this intentional

smrutilal2:


manolama:
Yup, definitely intentional as there are a lot of places where the log could be updated so we have to maintain the thread safety.

PR #2143, opened 2021-06-30
COMMON:

manolama:
- Fix a parsing bug of the CN in the RefreshingSSLContext and allow
  it to look at the extended KeyUsage fields for the serverAuth field.

DIST:
- Add the mock meta to the default dev configs.


abuthahir80 (approved):
👍

PR #2160, opened 2021-08-18
COMMON:

manolama:
- Add the nodeOption() method to QueryNodeConfig to support multiple options
  going forward to avoid adding a getter for each option. Also add a boolean
  getter for this option set.
- Add commonSourceConfig() to QueryPipelineContext as a helper to find out
  if a source needs IDs converted.
- Also add a downstreamQueryResultIds() method that will find out what IDs
  should be bubbling up to a node.
- Remove the padding bits from the TimeSeriesDataSourceConfig interface. Rename
  the timestamp getters. And add the getDataSource() method that will let
  sources override the ID, useful for pushdowns.
- Add the TimeAdjustment class and getAdjustments() to QueryPlanner that will
  take into account nodes that may require some query time padding or
  manipulation deviating from the query time.
- Add a baseSetupGraph() method to the QueryPlanner to make it easy for
  source implementations to setup their nodes in the graph.
- Add getRollupIntervals() to RollupConfig to fetch all of the applicable
  intervals. Was a private method before.
- Implement the ArrayAggregatorUtils to accumulate data into an array aggregator
  such as for joining time slices of time series.

CORE:
- Implement the above.
- Remove some leftover log lines from BaseAnomalyNode.
- Rework the HAClusterFactory to use the new routing utils and update it's
  config with less synchronization. It will no longer put in an HACluster node
  but add the re-worked Merger node. Behaves the same.
- Fixup some bugs in the DefaultQueryPlanner around pushdowns and ID converters.
  Also implement the time adjustments search.
- Rework the DownsampleFactory config for auto intervals to remove synchronization.
- Rework the Merger node to handle the HA use case and the split query use case.
  It now handles timeouts. Includes reworking the merger config of course.
- Add a numeric array iterator for the Merger to process split data.
  TODO - Need iterators for numeric and summary types too.
- Add some time validation to the RateFactory.
- Fix a bug in passthroughs for the Summarizer factory.
- Add the RoutingUtils class that handles HA or split queries and deals with
  pushdowns, rebuilding the graph if needed.
- Rework the TimeRouterEntry to account for time shifts and take an epoch in
  the matching methods to make unit testing cleaner. Entries now support
  timeouts for split queries.
- Rework TimeRouterFactory to support split queries. It has some configs around
  overlap and satisfying query requirements or being a bit lax.
- Tweak DefaultRollupConfig to return the default entry when needed.
- Add a new base query planner test class with lots of useful methods to test
  a graph being constructed properly.

HTTP:
- The BaseHttpExecutorFactory.nextHost() method will now return null if no
  healthy hosts were found instead of throwing an exception. that lets HA
  queries still run from another source if one is unable to find a healthy
  host. Also retry the call one more time for races in case the host was null.
- Add a temporary skip parsing flag to the V3 result for transitioning from an
  existing install to the newest code.

STORAGE:
- Use the data source config timestamps now instead of the query planner.


core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAggregatorUtils.java
```diff
@@ -1,2 +1,211 @@
...
+      ++arrayIndex;
+    }
+    if (aggIndex < 0) {
+      return AccumulateState.OUT_OF_BOUNDS;
+    }
+
+    int wrote = 0;
+    if (value.value().isInteger()) {
+      long[] array = value.value().longArray();
+      while (arrayIndex < value.value().end() &&
```
mailravi3390, line 124: This common part can be moved out of if else.

core/src/main/java/net/opentsdb/query/anomaly/BaseAnomalyNode.java
```diff
@@ -91,10 +118,13 @@ public BaseAnomalyNode(final QueryNodeFactory factory,
...
-      throw new IllegalStateException("Downsample can't be null.");
+      throw new IllegalStateException("Downsample can't be null for anomaly " +
+              "processing at this time.");
     }
-    final long query_time_span = context.query().endTime().msEpoch() - 
-        context.query().startTime().msEpoch();
+
+    dataSourceConfig = context.commonSourceConfig(this);
+    final long query_time_span = dataSourceConfig.endTimestamp().msEpoch() -
+            dataSourceConfig.startTimestamp().msEpoch();
```
mailravi3390, line 127: Why are we adding up start and end to get query time span ? Shouldnt it be end - start ?

core/src/main/java/net/opentsdb/query/anomaly/BaseAnomalyNode.java
```diff
@@ -113,8 +143,7 @@ public BaseAnomalyNode(final QueryNodeFactory factory,
     if (!Strings.isNullOrEmpty(config.getTrainingInterval())) {
```
mailravi3390, line 143: I think we need more logic to get the actual interval we need to give the data source, but that can come later I guess.

core/src/main/java/net/opentsdb/query/hacluster/HAClusterFactory.java
```diff
@@ -173,417 +124,135 @@ public void setupGraph(final QueryPipelineContext context,
...
-      if (!builder.dataSources().isEmpty()) {
-        if (planner.context().tsdb().getRegistry().getPlugin(
-              TimeSeriesDataSourceFactory.class, 
-              builder.dataSources().get(0)) == null) {
-          throw new IllegalArgumentException("No data source found for: " 
-            + builder.dataSources().get(0));
+    List<TimeRouterConfigEntry> supportedSources = null;
+    for (int i = 0; i < sources.size(); i++) {
+      final TimeRouterConfigEntry entry = sources.get(i);
+      if (!entry.factory().supportsQuery(context, config)) {
```
mailravi3390, line 150: Should the removal of the non-supported sources happen in the Query planner as opposed to here ?

core/src/main/java/net/opentsdb/query/hacluster/HAClusterFactory.java
```diff
@@ -173,417 +124,135 @@ public void setupGraph(final QueryPipelineContext context,
...
-          .setDataType(NumericType.TYPE.toString())
-          .build())
-        .setDataSource(config.getId())
-        .addSource(new_id)
-        .setId(config.getId())
-        .build();
+    MergerConfig merger = (MergerConfig) MergerConfig.newBuilder()
+            // TODO - want to make this configurable. Particularly when we're
+            // averaging or min-ing, etc.
+            .setAggregator("max")
```
mailravi3390, line 225: In case of getting data from multiple shards, this will have to be configurable (obtained from the group by) I guess.

core/src/main/java/net/opentsdb/query/plan/DefaultQueryPlanner.java
```diff
@@ -807,7 +851,98 @@ public String getMetricForDataSource(final QueryNodeConfig node,
...
+                  ((DownsampleConfig) predecessor).getInterval();
+          if (adjustments.downsampleInterval.equals(DownsampleConfig.AUTO)) {
+            final long deltaSeconds = context.query().endTime().epoch() -
+                    context.query().startTime().epoch();
+            adjustments.downsampleInterval =
+                    DownsampleFactory.getAutoInterval(context().tsdb(),
+                            deltaSeconds,
+                            ((DownsampleConfig) predecessor).getMinInterval());
+          } else if (((DownsampleConfig) predecessor).getRunAll()) {
+            adjustments.downsampleInterval = null;
```
mailravi3390, line 884: Does this mean there are multiple downsample intervals ? If not, we should consider the case where there are multiple downsample intervals.

core/src/main/java/net/opentsdb/query/plan/DefaultQueryPlanner.java
```diff
@@ -984,6 +1115,25 @@ protected void setupPushDowns() {
...
+          final QueryNodeConfig config = push_downs.get(i);
+          if (config.getSources().size() > 1 || config.resultIds().size() > 1) {
+            final String wanted = i > 0 ?
+                    push_downs.get(i - 1).getId() :
+                    node.getId();
+            final QueryNodeConfig.Builder builder = config.toBuilder();
+            builder.setSources(Lists.newArrayList(wanted))
+                    // TODO - WARNING - Gotcha here in that in the future if we
+                    // push down something like a summarizer -> gb -> ds we want
+                    // to serialize the summarizer && gb, we'd need to NOT remove
```
mailravi3390, line 1130: We do have a case like this in alerting for in_total case.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleFactory.java
```diff
@@ -288,10 +293,10 @@ public void setupGraph(final QueryPipelineContext context,
           new_source.addSummaryAggregation(newConfig.getAggregator());
         }
         plan.replace(source, new_source.build());
-      }      
+      }
+
+      plan.replace(config, newConfig);
```
mailravi3390, line 298: This seems to be replacing the given downsample config with the new one for every Tiemseries data source. In the end, we will choose the last seen Timeseries Data source config's start and end ?

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleFactory.java
```diff
@@ -257,29 +262,29 @@ public void setupGraph(final QueryPipelineContext context,
...
+              .setEnd(Long.toString(tsConfig.endTimestamp().msEpoch()))
+              .setId(config.getId())
+              .setResultIds(((DefaultQueryPlanner) plan).compileResultIds(config))
+              .build();
+
+      if (tsConfig.getSummaryAggregations() == null ||
+          tsConfig.getSummaryAggregations().isEmpty()) {
+        final TimeSeriesDataSourceConfig.Builder new_source =
+                (TimeSeriesDataSourceConfig.Builder) source.toBuilder();
         new_source.setSummaryInterval(newConfig.getInterval());
```
mailravi3390, line 288: Seems to me that newConfig.getInterval() is always going to be the default interval, given that we are not giving one in the builder.

core/src/main/java/net/opentsdb/query/processor/merge/Merger.java
```diff
@@ -81,54 +149,215 @@ public QueryNodeConfig config() {
...
+        final Waiter waiter = entry.getValue();
+        if (waiter.timeoutMillis > 0 && waiter.timeout == null) {
+          waiter.timeout = context.tsdb().getMaintenanceTimer().newTimeout(
+                  waiter, waiter.timeoutMillis, TimeUnit.MILLISECONDS);
+        }
+      }
+      return;
+    }
+
+    if (!complete.compareAndSet(false, true)) {
```
mailravi3390, line 223: Given that we already check complete.get() , is this needed ? Also, I assume the onNext method is called only by one thread at a time.

mailravi3390: My point is, if someone else other than this method is setting the complete variable to false on timeout.

core/src/main/java/net/opentsdb/query/processor/merge/MergerFactory.java
```diff
@@ -171,15 +200,31 @@ public TypedTimeSeriesIterator newIterator(final Merger node,
...
     @Override
     public TypedTimeSeriesIterator newIterator(final Merger node,
                                                final QueryResult result,
                                                final Map<String, TimeSeries> sources,
                                                final TypeToken<? extends TimeSeriesDataType> type) {
-      return new MergerNumericArrayIterator(node, result, sources);
+      switch (((MergerConfig) node.config()).getMode()) {
+        case HA:
+          return new MergerNumericArrayIterator(node, result, sources);
+        case SPLIT:
```
mailravi3390, line 222: If we dont implement the iterator for split, is the merge already done and the result represented as a single HA ?

core/src/main/java/net/opentsdb/data/types/numeric/aggregators/ArrayAggregatorUtils.java
manolama: I should test it but I'm trying to help tighten the loop for cache locality and avoid another branch. It could be compiled out at run time though, I'd have to see.

core/src/main/java/net/opentsdb/query/anomaly/BaseAnomalyNode.java
manolama: Yeah, for now it's user provided.

core/src/main/java/net/opentsdb/query/hacluster/HAClusterFactory.java
manolama: Not in this case cause it could be that one HA source is valid but the other isn't. The planner doesn't know if the other source is HA or split or what so it's a bit better here I think.

core/src/main/java/net/opentsdb/query/hacluster/HAClusterFactory.java
manolama: Yup

core/src/main/java/net/opentsdb/query/plan/DefaultQueryPlanner.java
manolama: Yeah, we only take the first one (predecessor) as that is what will affect the overlap in the data sources.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleFactory.java
manolama: This needs a little cleanup. I've moved the replace outside now but there are still some extra bits needed to handle multiple sources properly.

core/src/main/java/net/opentsdb/query/processor/downsample/DownsampleFactory.java
manolama: It actually computes it in the ctor (if it's auto, etc) which is why I'm leaving it for now. We can work out a better method to call a little later.

core/src/main/java/net/opentsdb/query/processor/merge/Merger.java
manolama: Yeah cause we have the race between the timeouts and data sources. The first `complete.get()` is just a double check to avoid doing any work building logs or starting timers in cases where the source finally comes in. The CAS is the final atomic gate. There's _still_ a potential to start extra timers but that should be pretty rare and would be caught by the CAS in the timeout.

core/src/main/java/net/opentsdb/query/processor/merge/MergerFactory.java
manolama: No, it would throw an NPE I believe. The Numeric and Summary will for now.

mailravi3390 (approved):
👍

PR #2161, opened 2021-08-23
Add drop_empty_tsuids query param to drop empty DPs

bshakur8:
Scenario: we have a db containing tags with high cardinalities. At a given timestamp values are set for metrics on a small subset of those tag values. When we query for data at a given time with wildcards for several of the tags, opentsdb first calculates the tsuids and then fills in the dps for each tsuid. However, in this case, most tsuids will be empty (contain no dps). The resulting json response is very large and is mostly filled with no data.  This change adds a flag drop_empty_tsuids to the query to allow opentsdb to filter tsuids from the results if there are no dps for a tsuid in the requested time period.

### DIFF
--- a/src/core/TSQuery.java
+++ b/src/core/TSQuery.java
@@ -108,6 +108,9 @@ public final class TSQuery {
   /** Override default max row count limit */
   private boolean override_data_point_limit;
   
+  /** Drop empty Data Points */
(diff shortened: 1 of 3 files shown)


balboah:
was there another fix for this? I'm having the same issue with results that has empty dps which I don't need to see

PR #2163, opened 2021-08-26
COMMON:

manolama:
- Add rehashing hack to the LongLongHashTable wherein we track the average
  number of scans (which will increase after deleting some data) and rehash
  when it hits a given threshold.
  TODO - configure that threshold.

### DIFF
--- a/common/src/main/java/net/opentsdb/collections/LongLongHashTable.java
+++ b/common/src/main/java/net/opentsdb/collections/LongLongHashTable.java
@@ -24,6 +24,12 @@ import java.math.BigInteger;
 /**
  * A linear probing primitive Map for long keys and long values. Stores data off heap.
  *
+ * NOTE: There is now a hacky, ugly way to rehash the map without resizing when
+ * deletes start to result in too many scans for missed entries. If the average
+ * number of scans per operation (any operation) exceeds the scan rehash threshold
+ * then we'll pick the next prime number from the primes set to hash with. It will
+ * roll over but by that time the key set should hopefully be fairly new.
+ *
  * @see DirectByteArray
  * @see LongLongIterator
  */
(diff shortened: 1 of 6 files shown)


common/src/main/java/net/opentsdb/collections/LongLongHashTable.java
```diff
@@ -43,17 +61,69 @@
...
+  private int prime;
+  private long scans;
+  private long opCount;
+  private double scanRehashThreshold;
+  private float growthRate;
+  private float growthDecay;
   private DirectByteArray table;
 
+  /**
+   * Default ctor with a 0.75 load, 0.75 growth factor, 0.75 growth decline and
```
smrutilal2, line 74: typo, default growth decay is `0.25`

common/src/main/java/net/opentsdb/collections/LongLongHashTable.java
```diff
@@ -78,16 +154,21 @@ public void close() {
...
 
   private void resize() {
-
+    long start = System.nanoTime();
+    int oldLength = arrayLength;
     int oldCap = slots;
     int newCap = oldCap;
     if (table != null) { // growing the table;
-      newCap = oldCap << 1; // double the size;
+      newCap = (int) (oldCap + (oldCap * growthRate));
```
smrutilal2, line 162: Now, after the table is grown, the new `size` may not be a prime. In such case the modulo hash `(key % size)` can cause collisions. That might be okay for now, but something to keep in mind.

common/src/main/java/net/opentsdb/collections/LongLongHashTable.java
```diff
@@ -112,9 +195,45 @@ private void resize() {
...
+    }
+    prime = PRIMES_FOR_HASHING[primeIndex];
+    scans = 0;
+    opCount = 0;
+
+    long oldAddress = table.init(arrayLength, false);
+    DirectByteArray oldTable = new DirectByteArray(oldAddress, false, slots * slotSz);
+    this.address = table.getAddress();
+    this.sz = 0;
+    if (oldTable != null) {
```
smrutilal2, line 217: redundant null check

smrutilal2 (commented):
Overall looks good. Just couple of typo and redundant null check

common/src/main/java/net/opentsdb/collections/LongLongHashTable.java
manolama: True, we have to keep an eye on that.

smrutilal2 (approved):

PR #2177, opened 2021-10-08
3x tweaks

manolama:


core/src/main/java/net/opentsdb/query/execution/serdes/JsonV2QuerySerdes.java
```diff
@@ -384,6 +384,60 @@ private void writeNumeric(
...
+          LOG.info("Out of RANGE!");
+          return;
+        }
+
+        break;
+      }
+      LOG.info("Found value: " + value);
+
+      while (timestamp.compare(Op.LT, end)) {
+        LOG.info("Writing ts: " + timestamp + " vs value:" + (value == null ? "null" : value.timestamp()));
```
gskchaitanya, line 411: Move this to debug?

manolama: ah damn need to remove it, thanks.

PR #2190, opened 2021-10-14
CORE:

manolama:
- Tweak HAClusterFactory to handle pushdowns in the config coming into
  setupQuery.
- Let HACluster check sources to see if they support pushdown.
- Fix a bug in the time adjustment search for DefaultQueryPlanner.
- Add downsample settings to the MergerConfig that we need when handling
  split queries with arrays.
- Fix the TimeSpec for MergerResult so that it reflects the split query time
  properly.

### DIFF
--- a/core/src/main/java/net/opentsdb/query/hacluster/HAClusterFactory.java
+++ b/core/src/main/java/net/opentsdb/query/hacluster/HAClusterFactory.java
@@ -171,10 +171,12 @@ public class HAClusterFactory extends BaseQueryNodeFactory<
     }
 
     if (sources.size() == 1) {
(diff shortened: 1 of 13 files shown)


core/src/main/java/net/opentsdb/query/hacluster/HAClusterFactory.java
```diff
@@ -333,7 +341,7 @@ public boolean supportsPushdown(
   @Override
   public RollupConfig rollupConfig() {
     // TODO - need to make sure downstream returns identical configs.
-    throw new UnsupportedOperationException();
+    return null;
```
anonthing, line 344: Shouldn't this still be Unsupported?

PR #2150, opened 2021-07-19
Run WriteToBuffer deferreds in chunks

ugtar:
fix OpenTSDB/opentsdb#839

In clusters containing tags with high cardinality a query over multiple tags can easily overflow the maximum size of the callback chain resulting in a CallbackOverflowError when creating the callback chain for writing the resulting json output. This change writes the json in chunks to avoid the overflow error.

PR #2204, opened 2021-11-03
Trigger callback chain when reached limit.

ugtar:
fix OpenTSDB#839

In clusters containing tags with high cardinality a query over multiple tags can easily overflow the maximum size of the callback chain resulting in a CallbackOverflowError when creating the callback chain for writing the resulting json output. This change writes the json in chunks to avoid the overflow error.

This PR supercedes https://github.com/OpenTSDB/opentsdb/pull/2150


RuffSeason:
Any plans to merge this? I recently built from this branch and I fixed tons of issues we were experiencing with querying high cardinality metrics.

PR #2205, opened 2021-11-12
PULSAR:

manolama:
- Add a basic Apache Pulsar consumer similar to the Kafka consumer.

### DIFF
--- a/storage/pulsar/src/main/java/net/opentsdb/transport/PulsarConsumerThread.java
+++ b/storage/pulsar/src/main/java/net/opentsdb/transport/PulsarConsumerThread.java
@@ -1,2 +1,84 @@
-package net.opentsdb.transport;public class PulsarConsumerThread {
+/*
+ * This file is part of OpenTSDB.
+ * Copyright (C) 2021  The OpenTSDB Authors.
+ *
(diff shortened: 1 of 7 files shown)


storage/pulsar/src/main/java/net/opentsdb/transport/PulsarConsumerThread.java
```diff
@@ -0,0 +1,84 @@
...
+      pulsarClient = new PulsarClientImpl(service.clientConfig());
+    } catch (PulsarClientException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+
+  @Override
+  public void run() {
+    try {
+      final Reader<byte[]> reader = pulsarClient.newReader()
```
gskchaitanya, line 50: NOTE: Consumer API has a reusable bytebuffer, we can move to Consumer API from this Reader API in the future.

gskchaitanya (approved):

storage/pulsar/src/main/java/net/opentsdb/transport/PulsarConsumerThread.java
manolama: Is that in the OSS client?

PR #2208, opened 2021-12-20
Bump Logback version to patch CVE-2021-42550

manolama:


SeanPMiller (approved):
Thanks. Log4shell has been fun. I confirmed that 1.2.9 does not appear vulnerable to CVE-2021-42550.

PR #2217, opened 2022-01-04
Enhance range tests and regex and add better default for output validator

### LINKED ISSUE #2211: Right axis broken on OpenTSDB UI
The fix for #2051 in #2127 introduced a new allow list for Gnuplot options under configuration property `tsd.gnuplot.options.allowlist`. If this property is not set then the `o` parameter is ignored. This is primarily used in the UI for controlling use of the right axis via value `axis x1y2`.

Testing locally has indicated that a value of `;axis x1y2` allows the UI to continue working with a mix of axis selections on the same Gnuplot render.

This should be set as the default value of this property.

eswdd:
Fixes #2211


manolama (approved):
Thanks!

PR #2219, opened 2022-01-05
V2 to V3 librange support

gskchaitanya:


core/src/main/java/net/opentsdb/query/pojo/TagVFilter.java
```diff
@@ -660,6 +660,14 @@ public TagVFilter build() {
             "The tagk cannot be null or empty");
       }
       
+      // V2 librange conversion
+      if (type.equals(TagVLiteralOrFilter.FILTER_NAME)) {
```
manolama, line 664: Doesn't Olympus actually set the `type` to `librange`? I thought it converted properly.

gskchaitanya: Olympus does set it to librange, but the API that's being used by the touchstone/deploy builds doesn't convert those.

core/src/main/java/net/opentsdb/query/pojo/TagVFilter.java
```diff
@@ -660,6 +660,14 @@ public TagVFilter build() {
...
+        if (filter.startsWith("librange://@")) {
```
anonthing, line 665: can this be handled in our plugin?

gskchaitanya: Yes, but it won't be clean. Several classes need to be extended and will introduce a bit more plugins as we discussed.

anonthing (approved):

PR #2226, opened 2022-02-02
upgrade log4j

smrutilal2:


pom.xml
```diff
@@ -125,14 +130,29 @@
...
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>jul-to-slf4j</artifactId>
-            <version>1.7.7</version>
+            <version>${slf4j.version}</version>
         </dependency>
-        
+        <dependency>
+          <groupId>org.apache.logging.log4j</groupId>
```
gskchaitanya, line 146: Are we overriding a transitive dependency here? Looks like we don't need log4j and if that is the case, can we exclude log4j from the dependency bringing the log4j?

smrutilal2: yes, this would set the version for the transitive dependencies

smrutilal2: I have excluded the transitive dependency. Please take a look

gskchaitanya (approved):

PR #2233, opened 2022-02-17
additionally build image for jdk15

smrutilal2:


manolama:
Might be worth just bumping to 17 for LTE. Also try building and running manually. I had problems when I tried to run it under 11.

smrutilal2:
> Might be worth just bumping to 17 for LTE. Also try building and running manually. I had problems when I tried to run it under 11.

Hi @manolama, the jdk8 container is leaking native memory. I have manually built and tested the container on jdk15. I wanted to move to jdk17 but I want to test it locally once.

smrutilal2:
And I had to use `-javaagent:/usr/share/opentsdb/lib/opentsdb-common-3.0.90-SNAPSHOT.jar` for the plugins to load correctly.

gskchaitanya (approved):
Looks good. Let’s give it a try

PR #2248, opened 2022-04-30
Distribution:

manolama:
- Create an all-in-one build with working Horizon Config to try out the UI.

### DIFF
--- a/distribution/src/resources/docker/opentsdb_all.yaml
+++ b/distribution/src/resources/docker/opentsdb_all.yaml
@@ -1,0 +1,93 @@
+tsd.network.port: 4242
+config.http.cache.directory: /tmp/
+tsd.http.staticroot: /usr/share/opentsdb/static
+
+tsd.http.async.timeout: 60000
+
+tsd.http.request.cors.pattern: .*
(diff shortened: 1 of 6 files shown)


distribution/src/resources/docker/opentsdb_all.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+  pluginLocations:
+  continueOnError: true
+  loadDefaultInstances: true
+
+# Mock Data Store Settings
+MockDataStore.register.writer: true
+
+tsd.query.tnrouter.config:
+  - sourceId: atlas
```
anonthing, line 64: Wanna remove this? 😁

anonthing (changes requested):

distribution/src/resources/docker/opentsdb_all.yaml
manolama: doh yeah thanks

anonthing:
Also I think having a aura only config would help for testing out OpenTSDB. Just all inmem aura for a day or so. For later though.

manolama:
Yeah I'll add that next and some code to fetch and populate test metrics.

PR #2272, opened 2023-04-10
Tighten up the regexes for Gnuplot URI params per multiple 

manolama:
security reports.

The best way of avoiding RCEs is to disable Gnuplot, but this should help a little.


oxeye-daniel:
Thanks @manolama , could you please open a GitHub security advisory for the vulnerability we found?

PR #2275, opened 2023-04-11
Improved fix for #2261.

manolama:
Regular expressions wouldn't catch the newlines or possibly other control characters. Now we'll use the TAG validation code to make sure the inputs are only plain ASCII printables first. Fixes CVE-2018-12972, CVE-2020-35476


oxeye-daniel:
@manolama - this seems to fix the [RCE we reported](https://github.com/OpenTSDB/opentsdb/issues/2261) before Synopsis on Dec 12, 2022. Seems like they have also issued a CVE for the vulnerability we found. Could you please open a Github security advisory? Thanks.

PR #2279, opened 2023-06-14
Fix XSS vulnerability for 3.0

MiInagaki:
Fixes for vulnerabilities similar to the following PRs have been made for the 3.0 branch.

- https://github.com/OpenTSDB/opentsdb/pull/2274
- https://github.com/OpenTSDB/opentsdb/pull/2275


manolama:
Thanks @MiInagaki. This code isn't live in the 3.x branch and should actually be removed. There are other UI's now so we don't need to rely on GnuPlot for 3.x.

MiInagaki:
Thanks @manolama. This PR simply reflects our internally maintained code, so we will not remove this code. 
We have detected vulnerabilities in the 3.x series in our system and we hope that they will be fixed in the future.

PR #2281, opened 2023-07-05
Test for null, avoid NPE

theultimatequestion:
fix OpenTSDB#2280

@manolama pinging you since you've committed the original. Hope that's alright.


manolama:
Thanks!

PR #2282, opened 2023-07-09
Use ConcurrentHashMap when serializing tags.

bshakur8:
fix OpenTSDB#1055

### DIFF
--- a/src/tsd/HttpJsonSerializer.java
+++ b/src/tsd/HttpJsonSerializer.java
@@ -22,6 +22,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.LinkedHashSet;
 import java.util.TreeMap;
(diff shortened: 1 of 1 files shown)


manolama:
Good catch, thanks!

ugtar:
I'm sorry @manolama, @bshakur8 changed this PR to a draft earlier today because we had a reproduction even with this patch. It may have reduced the frequency, but it's not a fix

manolama:
NP, thanks @ugtar it does help like you said. There can still be silent failures, I believe, in the tag resolutions deferreds. https://github.com/bshakur8/opentsdb/blob/78fb1fad8bc885a028bce6abdd2c4c2e62a99f29/src/tsd/HttpJsonSerializer.java#L916-L923 Try attaching some explicit error callbacks there.

